PR #7, opened 2017-04-05
Bug 1341730 - Add telemetry-batch-view docs

acmiyaguchi:
[Bug 1341730](https://bugzilla.mozilla.org/show_bug.cgi?id=1341730)

This adds most of the relevant documentation from telemetry-batch-veiw under datasets. These have not been edited for content. The only two documents that were not added are `analyzing_telemetry_data.md` and `E10SExperiment.md`.


harterrt:
It makes sense to merge these in as-is for now. We'll revise later. 

Thanks for getting this together, Anthony!

PR #34, opened 2017-07-21
Add 'new-profile' derived dataset to the docs

Dexterp37:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/firefox-data-docs/34)


Dexterp37:
Thanks @harterrt , that comment was very helpful! I think this is now ready for another look!

PR #39, opened 2017-08-09
Add Google Analytics

harterrt:


fbertsch:
Funny you just put this in. I was just thinking that if we're going to use GA for determining how much our sites are being used, we should ensure that our population of users don't all disable GA. I know I do!

harterrt:
Yep. I realized that I had privacy badger installed after too long debugging. I pushed this through to see if the plugin is even working - I can't seem to get anything to show up in GA.

If we don't see any traffic, we'll have to try something else. Otherwise, we should be able to use this as a directional measure.

PR #63, opened 2017-10-24
Add a note about opt-in data in the Longitudinal dataset.

Dexterp37:
This comes from [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1405239).

### DIFF
--- a/datasets/batch_view/longitudinal/intro.md
+++ b/datasets/batch_view/longitudinal/intro.md
@@ -24,6 +24,8 @@ In summary, the longitudinal table differs from `main_summary` in two important 
 * The longitudinal dataset groups all data so that one row represents a client_id
 * The longitudinal dataset samples to 1% of all client_ids
 
+Please note that this dataset only contains pre-release (or opt-out) histograms and scalars.
+
 #### Accessing the Data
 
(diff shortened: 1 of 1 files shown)


datasets/batch_view/longitudinal/intro.md
```diff
@@ -24,6 +24,8 @@ In summary, the longitudinal table differs from `main_summary` in two important
 * The longitudinal dataset groups all data so that one row represents a client_id
 * The longitudinal dataset samples to 1% of all client_ids
 
+Please note that this dataset only contains pre-release (or opt-in) histograms and scalars.
```
harterrt: Isn't the opposite true? We [only include opt-out histograms](https://github.com/mozilla/telemetry-batch-view/blob/master/src/main/scala/com/mozilla/telemetry/views/Longitudinal.scala#L111) in longitudinal?

Dexterp37: Of course you're right :)

mreid-moz:
Shouldn't this say "release channel (or opt-out)"?

Dexterp37:
Right. Fixed with #64

PR #64, opened 2017-10-24
Replace "pre-release" with "release" in the longitudinal intro

Dexterp37:


harterrt:
Passing the buck to @mreid-moz

PR #76, opened 2017-12-04
Update main_summary Schema

mreid-moz:


datasets/batch_view/main_summary/reference.md
```diff
@@ -92,6 +93,8 @@ root
  |-- env_build_version: string (nullable = true)
  |-- env_build_arch: string (nullable = true)
  |-- e10s_enabled: boolean (nullable = true)
+ |-- e10s_cohort: string (nullable = true)
```
harterrt: Are we really adding this field back?

mreid-moz: Heh, I think it re-appeared due to schema evolution. It certainly doesn't need to be in the docs though.

harterrt:
LGTM!

PR #77, opened 2017-12-18
Remove priority from bug template

harterrt:


harterrt:
5 month old bug for context: https://bugzilla.mozilla.org/show_bug.cgi?id=1379313

harterrt:
I'll merge tomorrow if I don't hear back.

PR #80, opened 2018-01-19
bug 1414839 - Document the public SSL dataset.

chutten:


datasets/other/ssl/intro.md
```diff
@@ -0,0 +1,22 @@
+The public SSL dataset publishes the percentage of pageloads Firefox users have performed
+that were conducted over SSL. This dataset is used to produce graphs like Let's Encrypt's
```
harterrt: Maybe link to the graph you reference here?

PR #88, opened 2018-02-16
Bug 1438337 - Initial documentation for error_aggregates dataset

wlach:


datasets/streaming/error_aggregates/reference.md
```diff
@@ -0,0 +1,93 @@
...
+## Sampling
+
+### Invalid Pings
+
+We ignore invalid pings in our processing. Invalid pings are defined as those that:
+
+* The submission dates or activity dates are invalid or missing.
+* The build ID is malformed.
+* The docType field is missing or unknown.
+* The ping does not include a positive count of usage hours.
```
wlach: @fbertsch could use some confirmation on this, I cribbed this section from crash_aggregates. I can easily see that the 3rd and 4th clauses are true from the source. The first and second are assumptions based on how we validate pings via the schema?

wlach:
To be clear on who I'm asking for what: 

* Asking @fbertsch for main review
* Asking @chutten mainly for review of the crash portions of this (but of course welcome feedback on anything)

datasets/streaming/error_aggregates/intro.md
```diff
@@ -0,0 +1,17 @@
...
+#### Contents
+
+The `error_aggregates` table contains counts of raw counts of various error measures,
+aggregated across each unique set of dimensions (for example: channel, operating system).
+You can get an aggregated count for any particular set of dimensions by summing using SQL.
+
+#### Accessing the data
+
+You can access the data via re:dash. Choose `Athena` and then select the
+`telemetry.error_aggregates` table. Here's an [example query](https://sql.telemetry.mozilla.org/queries/4769/source).
```
chutten: The chart in the example query is blank.

datasets/streaming/error_aggregates/reference.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Scheduling
+
+The `error_aggregates` job is run continously, using the spark streaming infrastructure
```
chutten: Is there a prod/docs link for our spark streaming infra?

datasets/streaming/error_aggregates/reference.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Schema
+
+The `error_aggregates` table has the following columns which define its dimensions:
+
+* `window_start`: Beginning of time when this sample was taken
+* `window_end`: End of time when this sample was taken (will always be 5 minutes more
```
chutten: I think window_* should us "interval" instead of "time" since it isn't an instant sample, it's over a period.

datasets/streaming/error_aggregates/reference.md
```diff
@@ -0,0 +1,93 @@
...
+  than `window_start` for any given row)
+* `channel`: the channel, like `release` or `beta`
+* `version`: the version e.g. `57.0.1`
+* `display_version`: like version, but includes beta number if applicable e.g. `57.0.1b4`
+* `build_id`: the YYYYMMDDhhmmss timestamp the program was built, like `20160123180541`. This is also known as the "build ID" or "buildid"
+* `application`: application name (e.g. `Firefox`)
+* `os_name`: name of the OS (e.g. `Darwin` or `Windows_NT`)
+* `os_version`: version of the OS
+* `country`: country code for the user (determined using geoIP), like `US` or `UK`
+* `experiment_id`: identifier of the experiment being participated in, such as `e10s-beta46-noapz@experiments.mozilla.org`, or null if no experiment
```
chutten: Should we comment that this should always be null since bug 1348748 introduced multiple-experiment annotations?

datasets/streaming/error_aggregates/reference.md
```diff
@@ -0,0 +1,93 @@
...
+* `experiment_branch`: the branch of the experiment being participated in, such as `control` or `experiment`, or null if no experiment
+* `e10s_enabled`: whether e10s is enabled
+* `gfx_compositor`: the graphics backend compositor used by the program, such as `d3d11`, `opengl` and `simple`. Null values may be reported as `none` as well
+* `quantum_ready`: whether the ping was submitted from an instance of Firefox that was Quantum-ready (i.e. no old-style add-ons)
+
+And these are the various measures we are counting:
+
+* `usage_hours`: number of user-hours represented by the aggregate
```
chutten: Do we want to mention that this is the sum of the lengths of the subsessions? Or that it is a measure of how long all the users in this aggregate had Firefox open? Or that it's 5min times the number of users who had an overlapping session?

datasets/streaming/error_aggregates/reference.md
```diff
@@ -0,0 +1,93 @@
...
+* `main_crashes`: number of main process crashes represented by the aggregate (or just program crashes, in the non-E10S case)
+* `content_crashes`: number of content process crashes represented by the aggregate
+* `plugin_crashes`: number of plugin process crashes represented by the aggregate
+* `gmplugin_crashes`: number of Gecko media plugin (often abbreviated GMPlugin) process crashes represented by the aggregate
+* `content_shutdown_crashes`: number of content process crashes that were caused by failure to shut down in a timely manner
```
chutten: Do we want to mention that we expect this number to always be less than content_crashes?

datasets/streaming/error_aggregates/reference.md
```diff
@@ -0,0 +1,93 @@
...
+* `gpu_crashes`: number of gpu process crashes
+* `browser_shim_usage_blocked`: number of times a CPOW shim was blocked from being created by browser code
+* `permissions_sql_corrupted`: number of times the permissions SQL error occurred (beta/nightly only)
+* `defective_permissions_sql_removed`: number of times there was a removal of defective permissions.sqlite (beta/nightly only)
+* `slow_script_notice_count`: number of times the slow script notice count was shown (beta/nightly only)
+* `
+The `error_aggregates` table has the following columns which define its dimensions:
```
chutten: Woops, some duplication here.

datasets/streaming/error_aggregates/intro.md
```diff
@@ -0,0 +1,17 @@
...
+
+#### Contents
+
+The `error_aggregates` table contains counts of raw counts of various error measures,
+aggregated across each unique set of dimensions (for example: channel, operating system).
+You can get an aggregated count for any particular set of dimensions by summing using SQL.
+
+#### Accessing the data
+
+You can access the data via re:dash. Choose `Athena` and then select the
```
fbertsch: When v2 lands it won't be available in Athena any longer, what with the hll columns.

datasets/streaming/error_aggregates/intro.md
```diff
@@ -0,0 +1,17 @@
+The `error_aggregates` table represents counts of errors counted from main and crash
+pings, aggregated every 5 minutes.
+
+#### Contents
+
+The `error_aggregates` table contains counts of raw counts of various error measures,
```
fbertsch: What are `error measures`? A few examples might help.

datasets/streaming/error_aggregates/reference.md
fbertsch: @wlach I'm not sure it's worth it to enumerate the various reasons we discard pings, the schema is pretty complex (and does include e.g. missing submission_date, malformed builds, etc.).

For this dataset in particular, we only process `main` and `crash` pings. Number 4 is true.

datasets/streaming/error_aggregates/reference.md
fbertsch: This was a bug in the code, it will be fixed in v2, which will have both `experiment_id` (for backwards compatibility) and `experiments`.

datasets/streaming/error_aggregates/reference.md
wlach: Yeah, I think you're right. It's probably just a background assumption that we discard this kind of invalid data, probably not really relevant to the intended audience here. I'll remove it.

datasets/streaming/error_aggregates/reference.md
wlach: Some variation on the second one ("a measure of how long all the users in this aggregate had Firefox open") seems like the most useful description for this audience, I'll make the change.

datasets/streaming/error_aggregates/intro.md
wlach: I'll delete the chart. I think that is a good example query in that it shows an example of getting a broad swathe of data, but maybe it would also be useful to do a more targetted query that shows something more graphable as well. I'll add that.

datasets/streaming/error_aggregates/reference.md
wlach: Not that I am aware of :/

datasets/streaming/error_aggregates/reference.md
wlach: ++ will fix

datasets/streaming/error_aggregates/reference.md
wlach: Sounds like a good idea

datasets/streaming/error_aggregates/intro.md
wlach: we need to think about whether that trade-off is worth it given the utility of the hll columns, as the performance of presto is pretty awful as discussed. anyway, i'll keep this as-is for now.

wlach:
@chutten can you have another look?

datasets/streaming/error_aggregates/reference.md
```diff
@@ -8,16 +8,54 @@
...
 
-We ignore invalid pings in our processing. Invalid pings are defined as those that:
+### Data sources
 
-* The submission dates or activity dates are invalid or missing.
-* The build ID is malformed.
-* The docType field is missing or unknown.
-* The ping does not include a positive count of usage hours.
+The aggregates in this data source are derived from main and crash [pings](../../pings.md).
+Crash pings are only used to count/gather main crash events, all other errors are gathered from main ones.
```
chutten: s/main ones/main pings/

wlach: I was trying not to overuse the word "pings", but it's probably less awkward your way (which was also my original wording). Will change.

PR #91, opened 2018-02-22
Add a quick mermaid.js example.

mreid-moz:


meta/contributing.md
```diff
@@ -70,6 +70,21 @@ Use **visual markup** like **bold text**, `code blocks`, and section headers.
 Avoid long paragraphs.
 Short paragraphs that describe one concept each makes finding important information easier.
 
+You may also use [mermaid.js](https://mermaidjs.github.io/) diagrams in code blocks:
```
harterrt: Do you mind breaking this out into it's own H2 header? I don't think it should be part of the style guide.

mreid-moz: Added a section for "Plugins" with mermaid as a sub-section. RFAL.

PR #112, opened 2018-03-30
Bug 1412875 - Add documentation for first_shutdown_summary

acmiyaguchi:
Documentation for the `first_shutdown_summary`.


datasets/batch_view/first_shutdown_summary/intro.md
```diff
@@ -0,0 +1,30 @@
...
+
+#### Background and Caveats
+
+Data collection latency was vastly improved with the introduction of the
+shutdown ping-sender mechanism in Firefox 55. To maintain consistent historical
+behavior, the first main ping is not sent until the second start up. Instead, a
+separate first-shutdown ping was created to maintain backwards compatibility.
+
+In many cases, the first-shutdown ping is a duplicate of the main ping. To
+avoid duplicate documents, take the `outer join` of the two datasets on
```
fbertsch: I'm not quite sure this is correct. Taking `main_summary m OUTER JOIN first_shutdown f ON m.doc_id = f.doc_id` will result in a dataset that only contains pings that were sent in either one, but not both. This should instead identify a query that gives the full `main_summary` schema, includes all rows, but not duplicates. You can achieve this with a `UNION ALL` in conjunction with a `LEFT OUTER JOIN`.

datasets/batch_view/first_shutdown_summary/intro.md
```diff
@@ -0,0 +1,30 @@
...
+their `document_id`.
+
+#### Accessing the Data
```
fbertsch: Is this dataset backfilled? If not how far back is the data available?

datasets/batch_view/first_shutdown_summary/intro.md
acmiyaguchi: The correct join would be a `FULL OUTER JOIN`, which would give the documents in both. I did some double checking in a notebook because I was worried about duplicate documents during the join.

It looks like the  `UNION ALL` and distinct [beats out any combination](https://dbc-caf9527b-e073.cloud.databricks.com/#notebook/9864/command/9877) of de-duplication and `FULL OUTER JOIN`.

datasets/batch_view/first_shutdown_summary/intro.md
acmiyaguchi: It's backfilled to 20170922, but it's probably most useful from Firefox 57 beta onward.

acmiyaguchi:
@fbertsch ready for another look

acmiyaguchi:
Thanks for the review!

PR #116, opened 2018-04-10
Clarify usage_hours in error aggregates

wlach:


harterrt:
Is this PR still active?

wlach:
Yup, I'll ping @fbertsch about it.

wlach:
Note to self: we may need to update this for `_v2`, usage hours may have a different meaning on fennec/android. But this seems like a good incremental improvement if I do say so myself.

PR #118, opened 2018-04-13
Tweak tone of summary towards product data

georgf:
Reading the summary casually, it sounds to me like we are focused on user data.
What we're really aiming for (and should communicate clearly) is that we're interested in understanding our products and their usage better.

I'm trying to slightly tweak the tone here.
Thoughts?

### DIFF
--- a/introduction.md
+++ b/introduction.md
@@ -1,7 +1,7 @@
 # Firefox Data Documentation
 
 This document will teach you how to use Firefox data
-to answer questions about our products and their users. The source for this documentation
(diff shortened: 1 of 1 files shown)


introduction.md
```diff
@@ -1,7 +1,7 @@
 # Firefox Data Documentation
 
 This document will teach you how to use Firefox data
-to answer questions about our users. The source for this documentation
+to answer questions about our products and their users. The source for this documentation
```
harterrt: "questions about how users interact with our products"

PR #119, opened 2018-04-13
Link to client custom ping cookbook

georgf:
Ideally we should have a single workflow, diving into both cookbooks, but this is a start.


cookbooks/new_ping.md
```diff
@@ -10,6 +10,8 @@ answer. General questions about "How do users use our product?" won't cut it - t
 specific, concrete asks that can be translated to data points. This will also make it easier down
 the line as you start data review.
 
+More detail on how to design and implement new pings [can be found here](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/collection/custom-pings.html).
```
fbertsch: This should specify that it is specific to Firefox Desktop telemetry.

PR #125, opened 2018-04-21
Add a keyed scalar example for longitudinal

jryans:
@chutten helped me work out this keyed scalar example on IRC. Let's add it to the docs as an example for others.


badboy:
It should at least include a short heading on what this query returns.

harterrt:
LGTM, @chutten can you review the query/description for accuracy?

cookbooks/longitudinal_examples.md
```diff
@@ -170,6 +170,22 @@ WITH samples AS
 SELECT approx_distinct(client_id) FROM samples WHERE max_concurrent_tabs > 100
 ```
 
+#### Keyed scalars
+
+Retrieve all the keys for a given scalar and sum all values for each key giving
+one row per key:
```
chutten: Should we remove the channel groupby and filter as it is incidental to this recipe?

I think it's pretty neat (I've never seen that interactive filter before!) so we could alternatively add it to the description.

chutten:
Oh, and though the query works with the version of presto we have on the "Presto" datasource, it doesn't work with the ancient one we have to put up with on the "Athena" datasource, so we should probably note that.

chutten:
@jryans You good to make these changes?

jryans:
Ah yeah, I was out for a few days and didn't notice the updates here. 😅 I'll make the changes noted.

jryans:
Okay, should be ready for another review @chutten.

harterrt:
Re:dash is the preferred spelling and shouldn't trigger an error

PR #131, opened 2018-05-07
activedau cookbook

SuYoungHong:
added documentation activeDAU under "cookbooks". 

not sure if this is the correct location for it.


cookbooks/active_dau.md
```diff
@@ -0,0 +1,30 @@
+# Active DAU
+
+An "Active User" is defined as a client who has total_daily_uri >= 5 URI for a given date.
```
harterrt: Please mark total_daily_uri as code e.g. `total_daily_uri`. This will prevent any issues with the spelling CI.

cookbooks/active_dau.md
```diff
@@ -0,0 +1,30 @@
+# Active DAU
+
+An "Active User" is defined as a client who has total_daily_uri >= 5 URI for a given date. 
+* Dates are defined by `submission_date_s3`.
+* A client's total_daily_uri is defined as their sum of `scalar_parent_browser_engagement_total_uri_count` for a given date[1].
```
harterrt: As above, please use backticks for total_daily_uri

cookbooks/active_dau.md
```diff
@@ -0,0 +1,30 @@
+# Active DAU
+
+An "Active User" is defined as a client who has total_daily_uri >= 5 URI for a given date. 
+* Dates are defined by `submission_date_s3`.
+* A client's total_daily_uri is defined as their sum of `scalar_parent_browser_engagement_total_uri_count` for a given date[1]. 
+
+Here is an example query in `main_summary` for getting Active DAU (aDAU) over time:
```
harterrt: DAU and aDAU are both triggering spelling errors, but should not be. Can you please add these terms to the .spelling file in the root of the repository?

cookbooks/active_dau.md
```diff
@@ -0,0 +1,30 @@
...
+            1, 2) as daily_clients_table
+WHERE
+    total_daily_uri >= 5
+GROUP BY 
+    1
+ORDER BY 
+    1 ASC
+```
+
+[1] Note, the probe measuring `scalar_parent_browser_engagement_total_uri_count` only exists in clients with FF50 and up. Clients on earlier versions of Firefox won't be counted as an Active User (regardless of their use). Similarly, `scalar_parent_browser_engagement_total_uri_count` doesn't increment when a client is in Private Browsing mode, so that won't be included as well.
```
harterrt: s/FF50/Firefox 50/

cookbooks/active_dau.md
```diff
@@ -0,0 +1,30 @@
...
+SELECT
+    submission_date_s3, 
+    count(DISTINCT client_id) as aDAU
+FROM 
+    (SELECT 
+            submission_date_s3, 
+            client_id, 
+            sum(coalesce(scalar_parent_browser_engagement_total_uri_count, 0)) as total_daily_uri
+        FROM
+            main_summary
```
harterrt: I'm concerned readers will blindly copy/paste this query and bork re:dash. Can you restrict the `submission_date_s3` and `sample_id` range?

cookbooks/active_dau.md
```diff
@@ -0,0 +1,30 @@
+# Active DAU
```
harterrt: Be sure to link to this document from `SUMMARY.md` or it will not be rendered.

cookbooks/active_dau.md
sunahsuh: `clients_daily_v6` is out as of last week and this is already a pre-computed field in the dataset: https://github.com/mozilla/telemetry-batch-view/blob/master/src/main/scala/com/mozilla/telemetry/views/ClientsDailyView.scala#L218

I'd encourage using that dataset over main_summary

SuYoungHong:
fixed issues and added a query for clients_daily_v6 query!

harterrt:
Woot!

PR #132, opened 2018-05-07
Update documentation for v2 of error_aggregates

akkomar:
This updates `error_aggregates`' docs to match v2 of the job. Clearly named columns are left without further explanations.

This should land after v2 version of the job is productionized.


akkomar:
@wlach please take a look - this is updated for recently deployed `v2`.

PR #134, opened 2018-05-15
Bug 1459662 - deprecate client_count

### LINKED ISSUE #129: Replace client_count with client_count_daily in Choosing a Dataset
https://docs.telemetry.mozilla.org/concepts/choosing_a_dataset.html

relud:
resolves #129


relud:
removing review, just found another thing i need to add will re-request review when ready

PR #152, opened 2018-06-15
Bug 1453376 - add 'Other Datasets' heading and hgpush docs

mars-f:
* Add a new heading to the TOC for 'Other Datasets'.
* Document the hgpush dataset under the new heading.


mars-f:
ping @harterrt

PR #153, opened 2018-06-19
Add retention cookbook

benmiroglio:
This is a cookbook that guides the reader through retention analysis. This is the most recent draft after an informal review by the majority of the Product Data Science Team.

### DIFF
--- a/cookbooks/retention.md
+++ b/cookbooks/retention.md
@@ -1,0 +1,211 @@
+*Authored by the Product Data Science Team. Please direct questions/concerns to Ben Miroglio (bmiroglio).*
+
+# Retention
+
+Retention measures the rate at which users are *continuing* to use Firefox, making it one of the more important metrics we track. We commonly measure retention between releases, experiment cohorts, and various Firefox subpopulations to better understand how a change to the user experience or use of a specific feature affect behavior.
+
+### N Week Retention
+
+Time is an embedded component of retention. Most retention analysis starts with some anchor, or action that is associated with a date (experiment enrollment date, profile creation date, button clicked on date *d*, etc.). We then look 1, 2, …, N weeks beyond the anchor to see what percent of users have submitted a ping (signaling their continued use of Firefox).
+
+For example, let’s say we are calculating retention for new Firefox users. Each user can then be anchored by their `profile_creation_date`, and we can count the number of users who submitted a ping between 7-13 days after profile creation (1 Week retention), 14-20 days after profile creation (2 Week Retention), etc.
+
+#### Example Methodology
+
+
+
+Given a dataset in Spark, we can construct a field `retention_period` that uses `submission_date_s3` to determine the period to which a ping belongs (i.e. if a user created their profile on April 1st, all pings submitted between April 8th and April 14th are assigned to week 1). 1-week retention can then be simplified to the percent of users with a 1 value for `retention_period`, 2-week retention simplifies to the percent of users with a 2 value for `retention_period`, ..., etc. Note that each retention period is independent of the others, so it is possible to have higher 2-week retention than 1-week retention (especially during holidays). 
+
+First let's map 1, 2, ..., N week retention the the amount of days elapsed after the anchor point:
+
+```python
+PERIODS = {}
+N_WEEKS = 6
+for i in range(1, N_WEEKS + 1):
+    PERIODS[i] = {
+        'start': i * 7,
+        'end': i * 7 + 6
+    }  
+```
+Which gives us
+
+```python
+{1: {'end': 13, 'start': 7},
+ 2: {'end': 20, 'start': 14},
+ 3: {'end': 27, 'start': 21},
+ 4: {'end': 34, 'start': 28},
+ 5: {'end': 41, 'start': 35},
+ 6: {'end': 48, 'start': 42}}
+
+```
+
+Next, let's define some helper functions: 
+
+```python
+import datetime as dt
+import pandas as pd
+import pyspark.sql.types as st
+import pyspark.sql.functions as F
+
+udf = F.udf
+
+def date_diff(d1, d2, fmt='%Y%m%d'):
+    """
+    Returns days elapsed from d2 to d1 as an integer
+    
+    Params:
+    d1 (str)
+    d2 (str)
+    fmt (str): format of d1 and d2 (must be the same)
+    
+    >>> date_diff('20170205', '20170201')
+    4
+    
+    >>> date_diff('20170201', '20170205)
+    -4
+    """
+    try:
+        return (pd.to_datetime(d1, format=fmt) - 
+                pd.to_datetime(d2, format=fmt)).days
+    except:
+        return None
+    
+
+@udf(returnType=st.IntegerType())
+def get_period(anchor, submission_date_s3):
+    """
+    Given an anchor and a submission_date_s3,
+    returns what period a ping belongs to. This 
+    is a spark UDF.
+    
+    Params:
+    anchor (col): anchor date
+    submission_date_s3 (col): a ping's submission_date to s3
+    
+    Global:
+    PERIODS (dict): defined globally based on n-week method
+    
+    Returns an integer indicating the retention period
+    """
+    if anchor is not None:
+        diff = date_diff(submission_date_s3, anchor)
+        if diff >= 7: # exclude first 7 days
+            for period in sorted(PERIODS):
+                if diff <= PERIODS[period]['end']:
+                    return period
+
+@udf(returnType=st.StringType())
+def from_unixtime_handler(ut):
+    """
+    Converts unix time (in days) to a string in %Y%m%d format.
+    This is spark UDF.
+    
+    Params:
+    ut (int): unix time in days
+    
+    Returns a date as a string if it is parsable by datetime, otherwise None
+    """
+    if ut is not None:
+        try:
+            return (dt.datetime.fromtimestamp(ut * 24 * 60 * 60).strftime("%Y%m%d"))
+        except:
+            return None
+        
+```
+
+Now we can load in a subset of `main_summary` and construct the necessary fields for retention calculations:
+
+```python
+ms = spark.sql("""
+    select client_id, 
+           submission_date_s3,
+           profile_creation_date,
+           os
+    from main_summary 
+    where submission_date_s3 >= '20180401'
+    and submission_date_s3 <= '20180603'
+    and sample_id = '42'
+    and app_name = 'Firefox'
+    and normalized_channel = 'release'
+    and os in ('Darwin', 'Windows_NT', 'Linux')
+    """)
+
+PCD_CUTS = ('20180401', '20180415')
+
+ms = (
+    ms.withColumn("pcd", from_unixtime_handler("profile_creation_date")) # i.e. 17500 -> '20171130'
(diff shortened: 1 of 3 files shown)


cookbooks/retention.md
```diff
@@ -0,0 +1,211 @@
+*Authored by the Product Data Science Team. Please direct questions/concerns to Ben Miroglio (bmiroglio).*
```
mreid-moz: drive-by: you'll have to add `Miroglio` to the `.spelling` file (similar with `interpretability`). You can either also put `bmiroglio` in the dictionary or surround it with backticks.

benmiroglio: Thanks for the guidance--Fixed!

cookbooks/retention.md
```diff
@@ -0,0 +1,211 @@
...
+
+Retention measures the rate at which users are *continuing* to use Firefox, making it one of the more important metrics we track. We commonly measure retention between releases, experiment cohorts, and various Firefox subpopulations to better understand how a change to the user experience or use of a specific feature affect behavior.
+
+### N Week Retention
+
+Time is an embedded component of retention. Most retention analysis starts with some anchor, or action that is associated with a date (experiment enrollment date, profile creation date, button clicked on date *d*, etc.). We then look 1, 2, …, N weeks beyond the anchor to see what percent of users have submitted a ping (signaling their continued use of Firefox).
+
+For example, let’s say we are calculating retention for new Firefox users. Each user can then be anchored by their `profile_creation_date`, and we can count the number of users who submitted a ping between 7-13 days after profile creation (1 Week retention), 14-20 days after profile creation (2 Week Retention), etc.
+
+#### Example Methodology
```
fbertsch: Would it be too difficult to add a pure SQL example?

benmiroglio: In addition to or in place of the current example?

fbertsch: I think in addition to would be fine, the more examples the merrier (IMO).

benmiroglio: Agreed :). I'd add one in.

cookbooks/retention.md
```diff
@@ -0,0 +1,211 @@
...
+            return (dt.datetime.fromtimestamp(ut * 24 * 60 * 60).strftime("%Y%m%d"))
+        except:
+            return None
+        
+```
+
+Now we can load in a subset of `main_summary` and construct the necessary fields for retention calculations:
+
+```python
+ms = spark.sql("""
```
harterrt: Nit: this SQL statement should follow the [style guide](https://docs.telemetry.mozilla.org/concepts/sql_style.html). In particular, please:

* Capitalize reserved words like `WHERE`, `SELECT`, and `AND` ([link](https://docs.telemetry.mozilla.org/concepts/sql_style.html#reserved-words))
* Add a newline after root keywords like `SELECT` and `WHERE` ([link](https://docs.telemetry.mozilla.org/concepts/sql_style.html#code-blocks))
* Indent lines 126-130 one additional level to clarify the `WHERE` statement block

cookbooks/retention.md
```diff
@@ -0,0 +1,211 @@
...
+
+When performing retention analysis between two or more groups, it is important to look at other usage metrics to get an understanding of other influential factors.
+
+For example, imagine you are tasked with reporting retention numbers for users that enabled sync (`sync_configured = true`) compared to users that haven't. You find that users with and without sync have a 1 week retention of 0.80 and 0.40, respectively. Wow--we should really be be promoting sync as it could double retention numbers!
+
+*Not quite*. Turns out you next look at `active_ticks` and `total_uri_count` and find that sync users report much higher numbers for these measures as well. Now how can we explain this difference in retention?
+
+There could be an entirely separate cookbook devoted to answering this question, however this contrived example is meant to demonstrate that simply comparing retention numbers between two groups isn't capturing the full story. Sans an experiment or model-based approach, all we can say is "enabling sync is **associated** with higher retention numbers." There is still value in this assertion, however it should be stressed that association/correlation != causation!
```
harterrt: Great explaination and example. It's worth bolding **correlation != causation** so the reader can get the gist of this example while scanning.

benmiroglio: good point!

benmiroglio:
I plan to add a Pure SQL example per @fbertsch 's request, however I think this is ok to merge for the moment after getting @SuYoungHong 's review.

cookbooks/retention.md
```diff
@@ -0,0 +1,211 @@
...
+        if diff >= 7: # exclude first 7 days
+            for period in sorted(PERIODS):
+                if diff <= PERIODS[period]['end']:
+                    return period
+
+@udf(returnType=st.StringType())
+def from_unixtime_handler(ut):
+    """
+    Converts unix time (in days) to a string in %Y%m%d format.
+    This is spark UDF.
```
jklukas: Nit: This is _a_ spark UDF

cookbooks/retention.md
```diff
@@ -0,0 +1,214 @@
+*Authored by the Product Data Science Team. Please direct questions/concerns to Ben Miroglio (`bmiroglio`).*
+
+# Retention
+
+Retention measures the rate at which users are *continuing* to use Firefox, making it one of the more important metrics we track. We commonly measure retention between releases, experiment cohorts, and various Firefox subpopulations to better understand how a change to the user experience or use of a specific feature affect behavior.
+
+### N Week Retention
```
SuYoungHong: good work! One more distinction I would make about types of N week retention: 

**rolling vs cohorted**

Here you're using a rolling approach to N week retention (N week is between (7*n) and ((7*n)+ 6) after client's pcd date), but a lot of people are using cohorts to calculate retention: 

Define weekly cohorts for your population (the clients that start between 20180101 - 20180107 = cohort1, 201801018- 20180114 = cohort2, etc.). For each cohort, define weekN retention as, appeared in the week period N weeks after the cohort period (i.e. week1 is appeared in 201801018- 20180114 for cohort 1)

So essentially, using an anchor **period** instead of an anchor **point**. 

Would be a good idea to clarify this distinction, since this is many people's understanding of N week retention right now.

cookbooks/retention.md
```diff
@@ -0,0 +1,214 @@
...
++----------+------+--------------+-------------+-------------------+
+|     Linux|     6|          1495|        22422|0.06667558647756668|
+|    Darwin|     6|          1288|         4734|0.27207435572454586|
+|Windows_NT|     6|         29024|       124872|0.23243000832852842|
++----------+------+--------------+-------------+-------------------+
+
+
+```
+
+we observe that 35.6% of Linux users whose profile was created in the first half of April submitted a ping 6 weeks later, and so forth. The example code snippets are consolidated in [this notebook](https://gist.github.com/benmiroglio/fc708e5905fad33b43adb9c90e38ebf4).
```
SuYoungHong: mentioned offline, but update copy to reflect code numbers

cookbooks/retention.md
benmiroglio: Good point. 

I use cohorts when there isn't a clear anchor point (i.e. users that enabled sync--there is no clear *sync was enabled on this date* indicator without looking across a client's history). Might be worth adding an example like this as well in section title "What if you don't have an achor point?". 

When you do have an anchor point however, IMO the rolling method is the most precise (and least expensive).

cookbooks/retention.md
benmiroglio: good catch! will fix.

cookbooks/retention.md
```diff
@@ -0,0 +1,214 @@
...
+# New vs. Existing User Retention
+
+The above example calculates **New User Retention**, which is distinct from **Existing User Retention**. This distinction is important when understanding retention baselines (i.e. does this number make sense?). Existing users typically have much higher retention numbers than new users. 
+
+Note that is more common in industry to refer to Existing User Retention as "Churn" (Churn = 1 - Retention), however, we use retention across the board for the sake of consistency and interpretability.
+
+**Please be sure to specify whether or not your retention analysis is for new or existing users.**
+
+
+
```
SuYoungHong: # Note on Point Anchoring vs Period Anchoring

As defined above, N week retention needs some anchor to start N from. 

However, we should note, the anchor itself can be a single point (such as a day) or a period (such as a week), also known as a cohort or acquisition period. 

The code example above is using a point anchor of a day, and defining the N week for each client independently, depending on which day they start. 

Many previous retention analysis at Mozilla, however, uses a period anchor, and calculates retention per cohort. For example: 

* Define a week (20180101 to 20180101) as the anchor/acquisition/cohort period. 
* Define clients who have `profile_creation_date` within these dates as a 'cohort'. 
* Define week Ns as 20180101 + 7N to 20180101 + 7N. 
* For the cohort, calculate each N retention. 
* Repeat for each cohort in the period of interest. 

The major difference between using a 1 week period anchor and a 1 day point anchor is that with the 1 week period anchor, the weekN retention period will be the same for a client who started at the beginning of the anchor period versus a client who started at the end of the anchor period. Whereas for point anchor, those same two clients will have their weekN retention period differ by the amount of days between when they started, respectively. 

**Point anchoring is preferred, however, both methods convey the same information.**

cookbooks/retention.md
```diff
@@ -0,0 +1,214 @@
...
+|    Darwin|     6|          1288|         4734|0.27207435572454586|
+|Windows_NT|     6|         29024|       124872|0.23243000832852842|
++----------+------+--------------+-------------+-------------------+
+
+
+```
+
+we observe that 35.6% of Linux users whose profile was created in the first half of April submitted a ping 6 weeks later, and so forth. The example code snippets are consolidated in [this notebook](https://gist.github.com/benmiroglio/fc708e5905fad33b43adb9c90e38ebf4).
+
+
```
acmiyaguchi: A couple of cleaning caveats that might be worth mentioning.

First is the client [submission latency](https://docs.telemetry.mozilla.org/datasets/mozetl/churn/reference.html#background-and-caveats), which can be between somewhere between 2-5 days. `clients_daily` accounts for this when aggregating clients per day, but it's not without it's own issues.

There are also noisy clients with inconsistent profile creation and sub-session dates. Different thresholds will affect the consistency of results. The cookbook could point to an existing baseline in the future.

benmiroglio: While I agree these are potential issues, I think it is beyond the scope of the cookbook since are general data issues subject to most Firefox analysis. I don't want to bog things down with all potential caveats, but rather serve as a jumping off point methodologically-speaking.

Since retention is usually calculated to compare two or more groups of users, I feel safe to exclude these caveats since each group is subject to the same pitfalls (in this case). 

@acmiyaguchi Let me know if you feel this is justified!

cookbooks/retention.md
benmiroglio: I'm working on a solution that uses views, as i don't think it is realistic to do everything in one giant query. More for the folks who are more SQL-inclined and not super familiar with Pyspark.

cookbooks/retention.md
acmiyaguchi: Yep, sounds good to me. A six week waiting period and relatively consistent results will probably take care of those caveats across analyses.

benmiroglio:
@SuYoungHong I wrote up a section that simplifies cohort-based retention, putting it in terms of the anchor point approach so the document is a little more cohesive. Please review!

SuYoungHong:
looks good to me! ready for merging!

harterrt:
:tada: :tada: :tada:

PR #155, opened 2018-07-05
Fix axis identification for evolution dashboard

nthomas-mozilla:
I noticed this while reading https://docs.telemetry.mozilla.org/concepts/analysis_intro.html#tmo


chutten:
Thanks for catching this!

PR #179, opened 2018-09-10
Bug 1488690 - Clarify experiment unpacking in error_aggregates

akkomar:


datasets/streaming/error_aggregates/intro.md
```diff
@@ -1,20 +1,27 @@
...
 unique set of dimensions (for example: channel, operating system) every 5
 minutes. You can get an aggregated count for any particular set of dimensions
 by summing using SQL.
 
+##### Experiment unpacking
+
+It's important to note that when this dataset is written, pings are "exploded" on (`experiment_id`, `branch_id`) and
+dimension with `null` (`experiment_id`, `branch_id`) is generated for each ping participating in any experiment.
+Therefore care must be taken when writing aggregating queries over the whole population - in these cases one needs to
+filter for `experiment_id=null and branch_id=null` in order to not double-count pings from experiments.
```
wlach: This paragraph confused me until I looked at your example below. I think this wording might be a little more clear:

"It's important to note that when this dataset is written, pings from clients participating in an experiment are aggregated on the `experiment_id` and `branch_id` dimensions corresponding to what experiment and branch they are participating. However, they are also aggregated with the rest of the population where the values of these dimensions are null. Therefore care must be taken when writing aggregating queries over the whole population - in these cases one needs to filter for `experiment_id=null and branch_id=null` in order to not double-count pings from experiments."

akkomar: Thank you, it is more clear indeed.

PR #164, opened 2018-07-19
WIP: new data: shield, delivery console

gregglind:

PR #181, opened 2018-09-12
Document SHIELD study datasets

tdsmith:
Here's a first pass at identifying datasets that are useful for analysts working on SHIELD studies.

This looks mostly complementary to #164 but some cross-references could be useful once either lands.

I'm conflicted about whether it would be useful and in-scope to provide more context or examples for how to e.g. look at enrollment or unenrollment events using the `events` table; what do y'all think?

### DIFF
--- a/datasets/shield.md
+++ b/datasets/shield.md
@@ -1,5 +1,121 @@
-# Work in Progress
+# Analyzing data from SHIELD studies
 
-This document is a work in progress.
-The work is being tracked in
-[this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1358286)
+This article introduces the datasets that are useful for analyzing SHIELD studies.
+After reading this article,
+you should understand how to answer questions about
+study enrollment,
+identify telemetry from clients enrolled in an experiment,
+and locate telemetry from add-on studies.
+
+## Table of contents
+
+<!-- toc -->
+
+## Dashboards
+
+[Firefox Test Tube](https://firefox-test-tube.herokuapp.com/)
+shows live enrollment activity and a selection of browser activity metrics
+for active pref-flip (but not add-on) experiments.
+
+## Experiment slugs
+
+Each experiment is associated with a slug,
+which is the label used to identify the experiment to Normandy clients.
+The slug is also used to identify the experiment in telemetry.
+
+You can determine the slug for a particular experiment by consulting
+[Test Tube](https://firefox-test-tube.herokuapp.com/) (only for pref-flip experiments),
+[this summary table](https://metrics.mozilla.com/~sguha/report/normandy_recipes.html),
+or the list of active recipes at
+https://normandy.cdn.mozilla.net/api/v1/recipe/signed/.
+
+## Tables
+
+These tables should be accessible from ATMO, Databricks, Presto, and Athena.
+
+### `experiments` column
+
+[`main_summary`](../datasets/batch_view/main_summary/reference.html),
+[`clients_daily`](../datasets/batch_view/clients_daily/reference.html),
+[`crash_summary`](../datasets/batch_view/crash_summary/reference.html),
+and some other tables
+include a `experiments` column
+which is a mapping from experiment stub to branch.
+
+You can collect rows from enrolled clients using query syntax like:
+
+```sql
+SELECT
+  *,
+  experiments['some-experiment-stub-12345'] AS branch
+FROM clients_daily
+WHERE experiments['some-experiment-stub-12345'] IS NOT NULL
+```
+
+### `experiments`
+
+The `experiments` table is a subset of rows from `main_summary`
+reflecting pings from clients that are currently enrolled in an experiment.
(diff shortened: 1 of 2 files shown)


datasets/shield.md
```diff
@@ -1,5 +1,121 @@
...
+
+## Table of contents
+
+<!-- toc -->
+
+## Dashboards
+
+[Firefox Test Tube](https://firefox-test-tube.herokuapp.com/)
+shows live enrollment activity and a selection of browser activity metrics
+for active pref-flip (but not add-on) experiments.
```
sunahsuh: That's.. not strictly true :/ We show add-on experiments if their normandy name matches with the experiment name that the addon registers as, which doesn't happen all the time.

datasets/shield.md
```diff
@@ -1,5 +1,121 @@
...
+### `events`
+
+The [`events` table](../datasets/batch_view/events/reference.html) includes
+Normandy enrollment and unenrollment events.
+
+Normandy events have event category `normandy`.
+The event value will contain the experiment slug.
+
+The event schema is described
+[in the Firefox source tree](https://hg.mozilla.org/mozilla-central/file/c4c125ee2556/toolkit/components/normandy/lib/TelemetryEvents.jsm#l12).
```
sunahsuh: Re: linking to a specific revision
Pro: the line number doesn't change
Con: the file could change and we'd have no idea that the link was out of date

Not sure what's better here, but I'm kind of inclined to link to latest and have people flag when it's pointing to the wrong thing?

datasets/shield.md
```diff
@@ -1,5 +1,121 @@
...
+[`mozilla-pipeline-schemas` repository](https://github.com/mozilla-services/mozilla-pipeline-schemas/tree/dev/schemas/telemetry/shield-study-addon).
+
+The key-value pairs are present in `data` attribute of the `payload` column.
+
+The `telemetry_shield_study_addon_parquet` table is produced by direct-to-parquet;
+data latency should be less than 1 hour.
+
+### `telemetry_shield_study_parquet`
+
+Telemetry from Normandy itself, containing enrollment and unenrollment events.
```
sunahsuh: Is it actually Normandy emitting these and not shield-study-utils?

datasets/shield.md
```diff
@@ -1,5 +1,121 @@
...
+These events also appear in the `events` table.
+
+The `telemetry_shield_study_parquet` table is produced by direct-to-parquet;
+data latency should be less than 1 hour.
+
+## Ping datasets
```
sunahsuh: Technically telemetry-cohorts a "source". I'd consider using the "Raw ping" terminology, which is commonly used for when we have to drop down to this level

datasets/shield.md
```diff
@@ -1,5 +1,121 @@
...
+
+## Dashboards
+
+[Firefox Test Tube](https://firefox-test-tube.herokuapp.com/)
+shows live enrollment activity and a selection of browser activity metrics
+for active pref-flip (but not add-on) experiments.
+
+## Experiment slugs
+
+Each experiment is associated with a slug,
```
sunahsuh: One normandy quirk is that every pref-flip experiment has a slug, while addon studies have a "name". Not sure why this is the case but so it goes.

datasets/shield.md
```diff
@@ -1,5 +1,121 @@
...
+which is the label used to identify the experiment to Normandy clients.
+The slug is also used to identify the experiment in telemetry.
+
+You can determine the slug for a particular experiment by consulting
+[Test Tube](https://firefox-test-tube.herokuapp.com/) (only for pref-flip experiments),
+[this summary table](https://metrics.mozilla.com/~sguha/report/normandy_recipes.html),
```
sunahsuh: There's also https://strategy-and-insights.mozilla.com/shield-studies/index.html

datasets/shield.md
tdsmith: Can I assert that add-ons have a slug specified by a `name` field?

datasets/shield.md
tdsmith: Added this and Experimenter to the dashboards section -- neither of these actually displays slugs!

datasets/shield.md
tdsmith: Thanks; I wasn't sure what to call them -- took another crack at this

datasets/shield.md
tdsmith: Linking the latest sgtm

datasets/shield.md
tdsmith: Added a caveat

datasets/shield.md
tdsmith: Whoops, you're right. Will Normandy still emit events that get collected to the `events` table for these?

tdsmith: Yes, it will.

tdsmith:
Great, thanks! I'll land this now and think about the common experiment queries cookbook.

PR #184, opened 2018-09-14
Clarify that document id is not optional

whd:
This has caused confusion and sadness on at least two occasions.


harterrt:
Is this ready to go?

PR #191, opened 2018-10-01
Bug 1489286 - Add some reference about channel behavior

fbertsch:

PR #192, opened 2018-10-01
Describe ping metadata struct

tdsmith:
First pass at https://bugzilla.mozilla.org/show_bug.cgi?id=1368133.

This attempts to give a point-in-time summary of the fields you might find in a ping today and doesn't attempt to call out the places where each of these fields are added. Does that feel all right?

(My motivation is that I keep forgetting how `geoCountry` is spelled :p)


datasets/pings.md
```diff
@@ -10,3 +10,48 @@
 
 You can find the reference documentation for all ping types
 [here](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/concepts/pings.html).
+
+# Ping Metadata
+
+The telemetry data pipeline appends a `meta` key to the ping body containing
```
mreid-moz: I would phrase this more generally, that the pipeline appends certain metadata fields containing extra (or sometimes repeated) data, and that the [Dataset API](https://python-moztelemetry.readthedocs.io/en/latest/api.html#dataset) makes it available as a `meta` key. These same fields are also commonly made available in direct-to-parquet datasets with a `metadata` prefix as described [here](https://github.com/mozilla/firefox-data-docs/pull/191/files#diff-8a341d077298aca144a3bd72201f1958R30).

datasets/pings.md
```diff
@@ -10,3 +10,48 @@
...
+`DNT` | Not present in all pings
+`docType` | e.g. "main"
+`documentId` | A UUID identifying the ping, generated by the client
+`geoCity` | from Geo-IP lookup of client IP; `??` if unknown
+`geoCountry` | from Geo-IP lookup of client IP; `??` if unknown
+`geoSubdivision1` | from Geo-IP lookup of client IP; not present in all pings
+`geoSubdivision2` | from Geo-IP lookup of client IP; not present in all pings
+`Host` | Public hostname of the ingestion endpoint
+`Hostname` | Private hostname of the ingestion endpoint
+`normalizedChannel` |
```
mreid-moz: This could also link to Frank's PR when it lands

datasets/pings.md
```diff
@@ -10,3 +10,48 @@
...
+`os` |
+`reason` | Documented [in the source tree](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/data/main-ping.html)
+`sampleId` | `crc32(clientId) % 100`
+`sourceName` | e.g. "telemetry"
+`sourceVersion` | Client `version` field (reflecting the version of the ping format)
+`submissionDate` | Server date (GMT) when ping was received, like `20180928`
+`telemetryEnabled` | Extracted value of `environment.settings.telemetryEnabled`, describing whether opt-in telemetry is enabled
+`Timestamp` | Server timestamp when the ping is received, expressed as nanoseconds since epoch
+`Type` | e.g. "telemetry"
+`X-PingSender-Version` | Not present in all pings
```
mreid-moz: Maybe link to more info about what `PingSender` is? Or just to [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1350044)?

trink:
Live documentation of the ingestion schemas for all data (the meta alteration happens down stream) can be found here: https://pipeline-cep.prod.mozaws.net/dashboard_output/analysis.moz_telemetry_parquet_schema.parquet.txt

tdsmith:
Thanks for the reviews!

I didn't feel strongly about linking to the parquet schema because I'm not sure how to interpret them and whether they affect my experience as a Dataset consumer and I'm selfishly interested in scratching my own itch -- if you think it'd be useful to document that in the text here I'm happy to open a followup PR.

PR #196, opened 2018-10-19
bug 1496775 - Document central Firefox Telemetry pings

chutten:
Through answering similar questions often enough we've found that there
is a need for a document spelling out, at a moderately-high leve, what are the
central pings we rely on to send us data, and how they behave.


chutten:
Throwing some mentions at @georgf and @badboy and @Dexterp37 for visibility (they helped review the draft)

PR #199, opened 2018-10-23
Add contact info page

sunahsuh:
This just came up in #dataplatform today, that we don't really have this info written down in an accessible place for folks outside the team


sunahsuh:
Fwiw the build failure is because of this bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1519945

.spelling
```diff
@@ -41,6 +41,7 @@ Databricks
 DataDog
 DataFrame
 DataFrames
+datapipeline
```
mreid-moz: drive-by comment: can you surround this and other additions to `.spelling` with backticks in the main text rather than adding them here? Including them as `#datapipeline` will avoid skip spell checking (similar for other slack channels / host names).

PR #202, opened 2018-11-02
Update python_moztelemetry documentation links

jklukas:
Adapting to new source of python_moztelemetry docs in https://github.com/mozilla/python_moztelemetry/pull/237


src/tools/spark.md
```diff
@@ -237,10 +237,9 @@ The `MozTelemetry` Library
 
 We have provided a library that gives easy access to the raw telemetry ping data.
 For example usage, see the
-[Telemetry Hello World](https://reports.telemetry.mozilla.org/post/tutorials/telemetry_hello_world.kp)
-example notebook.
-Detailed documentation for the library can be found at the
-[Python MozTelemetry Documentation](https://python-moztelemetry.readthedocs.io).
+[Telemetry Hello World example notebook](https://github.com/mozilla/mozilla-reports/blob/master/tutorials/telemetry_hello_world.kp/orig_src/Telemetry%20Hello%20World.ipynb).
```
jklukas: The kp article that renders the notebook is out of date with the notebook itself, so seems safer to link directly to the notebook (though this depends on GitHub rendering of notebooks, which is somewhat frequently offline).

src/tools/spark.md
```diff
@@ -237,10 +237,9 @@ The `MozTelemetry` Library
...
+Detailed API documentation for the library is linked from
+[the `python_moztelemetry` repository on GitHub](https://github.com/mozilla/python_moztelemetry).
```
wlach: IMO we should link to the actual docs, not the github repo.

src/tools/spark.md
wlach: sgtm

jklukas:
Changed link to hit the rendered docs rather than the GH repo

PR #225, opened 2018-11-06
Update event pipeline documentation

georgf:
Since the event ping landed, the pipeline looks a little bit different.
I've done some changes to update it.

### DIFF
--- a/src/concepts/pipeline/event_pipeline.md
+++ b/src/concepts/pipeline/event_pipeline.md
@@ -9,17 +9,18 @@ fx_code(fa:fa-cog Firefox code) --> firefox(fa:fa-firefox Firefox Telemetry)
 fx_extensions(fa:fa-cog Mozilla extensions) --> firefox
 fx_hybrid(fa:fa-cog Hybrid Content) --> firefox
 mobile(fa:fa-cog Mobile products) --> mobile_telemetry(fa:fa-firefox Mobile Telemetry)
-firefox -->|main ping| pipeline((fa:fa-database Firefox Data Pipeline))
-firefox -.->|events ping, planned| pipeline
+firefox -.->|main ping, Firefox <62| pipeline((fa:fa-database Firefox Data Pipeline))
+firefox -->|event ping, Firefox 62+| pipeline
 mobile_telemetry --> |mobile events ping| pipeline
-pipeline -->|Firefox events| main_summary[fa:fa-bars main summary table]
-main_summary --> events_table[fa:fa-bars events table]
+pipeline -->|Firefox <62 events| main_summary[fa:fa-bars main summary table]
+pipeline -->|Firefox 62+ events| events_table[fa:fa-bars events table]
+main_summary --> events_table
 pipeline -->|Mobile events| mobile_events_table[fa:fa-bars mobile events table]
 main_summary --> redash(fa:fa-bar-chart Redash)
 events_table --> redash
-events_table -.->|planned| amplitude(fa:fa-bar-chart Amplitude)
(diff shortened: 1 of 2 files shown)


georgf:
I believe the events dataset reference also needs updating.

src/concepts/pipeline/event_pipeline.md
```diff
@@ -90,16 +91,15 @@ use cases:
 - The [*dynamic event API*](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/collection/events.html#registerevents)
   allows code from Mozilla addons to record new events into Telemetry without shipping Firefox
   code.
-- The *Telemetry extension API* ([work in progress](https://bugzilla.mozilla.org/show_bug.cgi?id=1280234))
-  will allow Mozilla extensions to record new events into Telemetry.
+- The *[Telemetry WebExtension API](https://searchfox.org/mozilla-central/rev/55da592d85c2baf8d8818010c41d9738c97013d2/toolkit/components/extensions/schemas/telemetry.json#87)* ([introduced here](https://bugzilla.mozilla.org/show_bug.cgi?id=1280234))
```
chutten: There's no such thing as WebExtensions. Only "Extensions" and "Legacy Extensions"

georgf:
> I believe the events dataset reference also needs updating.
I've specifically left this out of this PR, because i don't know the details on this.
This is possibly a follow-up for Sunah or others?

src/concepts/pipeline/event_pipeline.md
```diff
@@ -9,17 +9,18 @@ fx_code(fa:fa-cog Firefox code) --> firefox(fa:fa-firefox Firefox Telemetry)
...
+pipeline -->|Firefox 62+ events| events_table[fa:fa-bars events table]
+main_summary --> events_table
 pipeline -->|Mobile events| mobile_events_table[fa:fa-bars mobile events table]
 main_summary --> redash(fa:fa-bar-chart Redash)
 events_table --> redash
-events_table -.->|planned| amplitude(fa:fa-bar-chart Amplitude)
+events_table -->|on request| amplitude(fa:fa-bar-chart Amplitude)
 mobile_events_table --> redash
-mobile_events_table --> amplitude
+mobile_events_table -->|on request| amplitude
```
sunahsuh: Technically, both the desktop and mobile events --> amplitude operate on heka-framed pings so there should just be one line, `pipeline -->|on request| amplitude

src/concepts/pipeline/event_pipeline.md
georgf: It's a confusing situation. We talk about "extensions"/"browser extensions" internally, but still call the API [the "WebExtension API" ](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions).
I think i'll just leave this as is for now, it should be clear enough i hope.

chutten: The title of that page is "Browser Extensions" :) 

It was a deliberate move around 57 or so when we were removing Legacy Extensions to specifically call the newer ones just Extensions. I would like to press this matter.

georgf: The page talks about the WebExtensions API though.

georgf: Also [here](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API).

georgf: Clarified that WebExtension API is a term still.
Clarified that there is a separate concern about not calling the extensions themselves WebExtensions.

PR #227, opened 2018-11-15
Add a cookbook for Normandy events

tdsmith:
Describe how to do some common counting operations on the Normandy events in the events table.

PR #233, opened 2018-12-01
Add an addon enrollment-count example

tdsmith:
Followup to #227.

### DIFF
--- a/src/cookbooks/normandy_events.md
+++ b/src/cookbooks/normandy_events.md
@@ -15,7 +15,7 @@ or name (for add-on experiments).
 Normandy events are described in detail in the
 [Firefox source tree docs][normandy-doc].
 
-Note that addon studies do not have branch information in the events table,
+Addon studies do not have branch information in the events table,
 since addons, not Normandy, are responsible for branch assignment.
 For studies built with the [add-on utilities][`addon-utils`],
 branch assignments are published to the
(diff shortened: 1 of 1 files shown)


tdsmith:
My attempt to dedupe client-events may not be correct!

It's necessary because [sometimes](https://github.com/mozilla/webcompat-blipz-experiment/issues/120) events are sent [more than once](https://github.com/mozilla/shield-cloudstorage/issues/41).

But StackOverflow [alleges](https://stackoverflow.com/questions/39505599/spark-dataframe-does-groupby-after-orderby-maintain-that-order?rq=1) that order is not stable over grouping.

An alternative would be to write something like

```python
def generate_row_with_min_index_udf(sdf, key_field):
  @f.pandas_udf(sdf.schema, f.PandasUDFType.GROUPED_MAP)
  def udf(df):
    return df.sort_values(key_field).fillna(method="bfill").iloc[:1,]
  return udf

my_events = (
  shield
  .filter(shield.submission >= EXPERIMENT_START)
  .filter(shield.payload.getItem("study_name") == STUDY_NAME)
  .filter(shield.payload.getItem("testing") == False)
  .withColumn("event", shield.payload.getItem("data").getItem("study_state"))
  .withColumn("branch", shield.payload.getItem("branch"))
  .select(shield.client_id, shield.submission, "event", "branch")
)

earliest_row = generate_row_with_min_index_udf(my_events, "submission")

events_by_branch = (
  my_events
  .groupBy(shield.client_id, "event")
  .apply(earliest_row)
  .groupBy("submission", "event", "branch")
  .agg(f.count("*").alias("n"))
)
```

which is not exceptionally performant and feels unwieldy, but maybe this is an unwieldy problem.

mreid-moz:
Another possibility is to use [`dropDuplicates`](https://spark.apache.org/docs/latest/api/python/pyspark.sql.html#pyspark.sql.DataFrame.dropDuplicates) using the two columns in question, but afaik it also doesn't preserve order - how important is it to get the first item?

Aside: my kingdom for broadly available [`DISTINCT ON`](https://www.postgresql.org/docs/11/sql-select.html#SQL-DISTINCT) functionality :-/

felixlawrence:
Would this be a scenario where you should `GROUP BY` everything you're selecting other than the timestamp, which you apply `F.min()` to?

If you need to group by only certain columns and pull out the data from other columns of the very first row in the group - e.g. a client enrolls in one branch then another then a third, and you want to find the branch they first enrolled in and ignore the other branches - then yes you'll need a different approach. Here's three:
1. Write a subquery that finds the minimum timestamp associated with each `(client_id, ...)` tuple that describes what you're wanting to group by. Then join this to the original table, filtering out the row(s) that match this timestamp.
2. Do a `LEFT JOIN` looking for older events, and filter out any events where the join succeeded.
3. Drop to an RDD, key it by what you're trying to group by, then `reduceByKey`. Example (where I was using `timestamp` to break a tie between `submission_date_s3`):
```
k1 = mse_rdd.keyBy(lambda x: x.client_id)
k2 = k1.reduceByKey(
    lambda x, y: 
        x if x.submission_date_s3 < y.submission_date_s3
        else y if y.submission_date_s3 < x.submission_date_s3
        else x if x.timestamp < y.timestamp else y
)
```

tdsmith:
> Would this be a scenario where you should GROUP BY everything you're selecting other than the timestamp, which you apply F.min() to?

Oops -- yes, I think it is. It looks like I was trying to count a client_id exactly once, even if the client_id was enrolled in more than one branch, and to extract the branch matching that earliest enrolment. I think that is not a useful requirement, so this approach seems both correct and simpler.

wlach:
@tdsmith is this still relevant? or should we close?

tdsmith:
I should just delete the pyspark sections; let me do that now.

tdsmith:
Oh, and this is all new; this can go away. Thanks!

PR #238, opened 2019-01-14
Document Heartbeat ping

tdsmith:
Closes https://bugzilla.mozilla.org/show_bug.cgi?id=1513044.

### DIFF
--- a/src/datasets/heartbeat.md
+++ b/src/datasets/heartbeat.md
@@ -1,6 +1,9 @@
 # Accessing Heartbeat data
 
 [Heartbeat][heartbeat] survey studies return telemetry on user engagement with the survey prompt.
+The heartbeat pings do not contain the survey responses themselves,
+which are stored by SurveyGizmo.
+
 The telemetry is received using the `heartbeat` document type,
 which is [described in the Firefox source tree docs][hbping].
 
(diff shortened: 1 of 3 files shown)


tdsmith:
hm, build is red because w.t.m.o is serving 503s

tdsmith:
PR for the broken link at #239.

src/datasets/heartbeat.md
```diff
@@ -0,0 +1,89 @@
...
+The `userid` is reported by appending it to the `surveyId` field of the ping, like:
+
+```
+hb-example-slug::e87bcae5-bb63-4829-822a-85ba41ee5d53
+```
+
+These can be extracted from the Parquet table for analysis using expressions like:
+
+```
+SPLIT(payload.survey_id,'::')[1] AS Shield_ID
```
harterrt: I'm confused by this. Is the `client_id` reported in `telemetry_heartbeat_parquet` not a telemetry `client_id`? If not, we should call this out explicitly in the introduction of the article. 

Otherwise, please clarify when one would want to extract the `shield_id`. I can't think of a reason offhand.

tdsmith: I should streamline the nomenclature -- `client_id` is a telemetry client_id; `userid`/`Shield_ID` is the random token given to SurveyGizmo. Knowing pairs of (`client_id`, `surveygizmo_userid`) lets you join survey responses to telemetry. I'll take another swing at it.

harterrt: Ah! Are the survey responses stored somewhere else? If so, I think that's the piece I was missing.

tdsmith: They are! I'll make sure that's clear.

tdsmith:
@harterrt Could you take another look?

tdsmith:
Thank you!

PR #252, opened 2019-02-20
Add a link to telemetry-next-node

mreid-moz:


mreid-moz:
Per IRC discussion, we want to highlight that `telemetry-next-node` is a legacy project, so I created a new section and moved this project there.

@fbertsch RFAL

PR #254, opened 2019-02-27
Add docs for the telementry_aggregates dataset

robhudson:


src/datasets/batch_view/telemetry_aggregates/reference.md
```diff
@@ -0,0 +1,91 @@
...
+    channel = 'nightly'
+    AND metric = 'GC_MS'
+    AND aggregate_type = 'build_id'
+    AND period = '201901'
+GROUP BY
+    submission_date
+ORDER BY
+    submission_date
+;
+```
```
robhudson: Open to idea for a better example SQL query here.

PR #255, opened 2019-03-21
Specify that DTMO uses CommonMark

tdsmith:
mdbook uses a CommonMark crate, so CommonMark it is.

Remove reference to AsciiDoc to avoid confusion.


src/meta/contributing.md
```diff
@@ -64,10 +64,8 @@ to check.
 
 ## Style Guide
 
-Articles should be written in
-[Markdown](https://daringfireball.net/projects/markdown/syntax)
-(not [AsciiDoc](http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/)).
-Markdown is usually powerful enough and is a more common technology than AsciiDoc.
+Articles should be written in Markdown—
```
badboy: Shouldn't there be a space before the dash?

badboy:
spellcheck complains about `CommonMark`

src/meta/contributing.md
```diff
@@ -64,10 +64,8 @@ to check.
...
+specifically, the [CommonMark dialect][commonmark].
```
mreid-moz:
```suggestion
specifically, the [`CommonMark` dialect][commonmark].
```

src/meta/contributing.md
tdsmith: It should be symmetric, anyway; I was hoping that ending a line with an em dash would suppress the whitespace from the newline, but it does not

src/meta/contributing.md
tdsmith: I went the .spelling route on the basis that it's a trademark like CircleCI or Django, which are both in .spelling, instead of a literal reference to a code entity

PR #257, opened 2019-03-25
Add documentation for clients_last_seen dataset

relud:


src/concepts/analysis_intro.md
```diff
@@ -50,4 +50,4 @@ Where to go next
 [measurement_dash]: https://telemetry.mozilla.org/new-pipeline/dist.html
 [evo_dash]: https://telemetry.mozilla.org/new-pipeline/evo.html
 [evo_gc_ms]: https://telemetry.mozilla.org/new-pipeline/evo.html#!aggregates=median!95th-percentile&cumulative=0&end_date=2017-06-13&keys=!__none__!__none__&max_channel_version=nightly%252F56&measure=GC_MS&min_channel_version=nightly%252F53&processType=*&product=Firefox&sanitize=1&sort_keys=submissions&start_date=2017-06-12&trim=1&use_submission_date=0
-[problem_client]: http://reports.telemetry.mozilla.org/post/projects%2Fproblematic_client.kp
+[problem_client]: http://mozilla.report/post/projects%2Fproblematic_client.kp/index.html
```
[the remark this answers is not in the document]
jklukas: FYI: https://bugzilla.mozilla.org/show_bug.cgi?id=1539155 for looking into whether we can relax the index.html requirement.

src/datasets/bigquery/clients_last_seen/intro.md
```diff
@@ -0,0 +1,41 @@
...
+User counts generated using `last_seen_date` can use `SUM` to reduce groups,
+because a given `client_id` will only be in one group per `submission_date`. So
+if MAU were calculated by `country` and `channel`, then the sum of the MAU for
+each `country` would be the same as if MAU were calculated only by `channel`.
+
+#### Accessing the Data
+
+The data is available in Re:dash and BigQuery.
+Take a look at this
+[example query](https://sql.telemetry.mozilla.org/queries/62029/source#159510).
```
jklukas:
```suggestion
[full running example query in Re:dash](https://sql.telemetry.mozilla.org/queries/62029/source#159510).
```

src/datasets/bigquery/clients_last_seen/reference.md
```diff
@@ -0,0 +1,73 @@
...
+# Data Reference
+
+## Example Queries
+
+#### Compute DAU for non-windows clients for the last week
+
+```sql
+SELECT
+    submission_date,
+    CURRENT_TIMESTAMP() AS generate_time,
```
jklukas: Why include a generated time in these examples? If this should stay, should it be `generated_timestamp` to be consistent?

src/concepts/analysis_intro.md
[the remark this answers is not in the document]
jklukas: Harold fixed the bug this morning, so it's no longer necessary to append index.html to these links.

src/datasets/bigquery/clients_last_seen/reference.md
relud: we consistently call it `generated_time` in bigquery-etl, so I think we should stick with that

I think if we use the field in bigquery-etl that we should use it in examples, so that people get used to seeing and including it as a general good practice. that said, I think we might want to drop the column in general as per mozilla/bigquery-etl#38

relud: removing the column

src/datasets/bigquery/clients_last_seen/intro.md
```diff
@@ -0,0 +1,41 @@
+The `clients_last_seen` dataset is useful for determining exact user counts
+such as MAU, WAU, and DAU.
+
+The `clients_last_seen` dataset is similar to the
+[`client_count_daily` dataset](/datasets/batch_view/client_count/reference.md)
+except it is used to efficiently generate exact counts, and it includes the most
+recent values in a 28 day window for all columns from the
```
harterrt: When would one prefer `client_count_daily` to `clients_last_seen`?

relud: `clients_last_seen` isn't available outside bigquery. Otherwise I can think of no reason not to deprecate `client_count_daily` in favor of `clients_last_seen`

relud:
DO NOT MERGE

the format of this table is on the verge of change, and I think it would be better to wait for that work

relud:
reflagging for review because I feel the changes since review are significant

src/cookbooks/active_dau.md
```diff
@@ -1,14 +1,14 @@
...
 
-**Active DAU** is the number of Active Users on a given day.
+**Active DAU** (aDAU) is the number of Active Users on a given day.
 
-**Active MAU** is the number of unique clients who have been an Active User on any day in the last **28 days**. In other words, any client that contributes to Active DAU in the last 28 days would also contribute to Active MAU for that day. Note that this is not simply the sum of Active DAU over 28 days, since any particular client could be active on many days.
+**Active MAU** (aMAU) is the number of unique clients who have been an Active User on any day in the last **28 days**. In other words, any client that contributes to aDAU in the last 28 days would also contribute to aMAU for that day. Note that this is not simply the sum of aDAU over 28 days, since any particular client could be active on many days.
 
-For quick analysis, using `clients_daily_v6` is recommended. Below is an example query for getting Active DAU (aDAU) using `clients_daily_v6`.
+For quick analysis, using `clients_daily_v6` is recommended. Below is an example query for getting aDAU using `clients_daily_v6`.
```
jklukas:
```suggestion
For quick analysis, it's possible to extract aDAU directly from `clients_daily_v6`, but note that calculating aMAU from the same data is significantly more complicated and expensive due to the need for a self-join over a 28 day window. Below is an example query for getting aDAU using `clients_daily_v6`.
```

src/cookbooks/active_dau.md
```diff
@@ -24,6 +24,21 @@ ORDER BY
     1 ASC
 ```
 
+[`clients_last_seen`](../datasets/bigquery/clients_last_seen/reference.md) can also be used, but is only available in BigQuery. Below is an example query for getting aDAU and aMAU using `clients_last_seen`.
```
jklukas:
```suggestion
[`clients_last_seen`](../datasets/bigquery/clients_last_seen/reference.md) is a more efficient option that includes precalculated aMAU, but is only available in BigQuery. Below is an example query for getting aDAU and aMAU using `clients_last_seen`.
```

src/cookbooks/active_dau.md
jklukas: If you think including this caveat is useful, it should also appear in the DAU/MAU doc.

src/cookbooks/active_dau.md
```diff
@@ -24,6 +24,21 @@ ORDER BY
...
+SELECT
+    submission_date,
+    COUNTIF(days_since_visited_5_uri < 1) AS visited_5_uri_dau,
+    COUNTIF(days_since_visited_5_uri < 28) AS visited_5_uri_mau
+FROM
+    telemetry.clients_last_seen
+GROUP BY
+    1
+ORDER BY
+    1 ASC
```
jklukas: I'd prefer to be verbose and name columns in the group by and order by clauses (so `submission_date` rather than 1) for docs and for production ETL.

src/datasets/bigquery/clients_last_seen/intro.md
```diff
@@ -0,0 +1,55 @@
...
+
+#### Content
+
+For each `submission_date` this dataset contains one row per `client_id`
+that appeared in `clients_daily` in a 28 day window including
+`submission_date` and preceding days.
+
+The `days_since_seen` column indicates the difference between `submission_date`
+and the most recent `submission_date` in `clients_daily` where the `client_id`
+appeared.
```
jklukas:
```suggestion
appeared. A client observed on the given `submission_date` will have `days_since_seen = 0`.
```

src/datasets/bigquery/clients_last_seen/intro.md
```diff
@@ -0,0 +1,55 @@
...
+values from `clients_daily` for each `client_id` in a 28 day window. This means
+[Active MAU](../../../cookbooks/active_dau.md)
+as defined cannot be efficiently calculated using `days_since_seen` because if
+a given `client_id` appeared every day in February and only on February 1st had
+`scalar_parent_browser_engagement_total_uri_count_sum >= 5` then it would only
+be counted on the 1st, and not the 2nd-28th. Active MAU can be efficiently and
+correctly calculated using `days_since_visited_5_uri`.
+
+MAU can be calculated over a `GROUP BY submission_date[, ...]` clause using
+`COUNT(*)`, because there is exactly one row in the dataset for each
```
jklukas: Perhaps it's worth pointing out that these three are equivalent:

```
count(days_since_seen < 28)
count(days_since_seen)
count(*)
```

for the specific case of calculating general MAU, but the most specific form is necessary to express both a different window and a different usage criterion.

src/cookbooks/active_dau.md
relud: I agree, but I wanted to match the existing style in the document

src/cookbooks/active_dau.md
relud: getting aDAU isn't more efficient in `clients_last_seen`, because `COUNT(*)` in `clients_daily` is v efficient. It is more efficient for aMAU, but sometimes aDAU is enough for people

relud:
@jklukas gret review, will address with follow-up PRs to avoid blocking this documentation any longer

PR #261, opened 2019-04-01
Recommend clients_last_seen for DAU and Active DAU

relud:
also indicate that `client_count_daily` cannot calculate Active MAU at this time.
also fix up some queries and mention how to use them in BigQuery.


src/cookbooks/dau.md
```diff
@@ -7,108 +7,76 @@ For the purposes of DAU, a profile is considered active if it sends any main pin
...
+  submission_date,
+  cardinality(merge(cast(hll AS HLL))) AS dau
 FROM
-  clients_daily_v6
+  telemetry.client_count_daily
 WHERE
-  submission_date_s3 <= '20181216'
-  AND submission_date_s3 >= '20181119'
-```
-Generating a table of MAU over time is more complex.  Here's one possible way, which is conceptually simple, but not very computationally efficient:
```
relud: cc @mreid-moz about removing these MAU examples in favor of `clients_last_seen`

PR #262, opened 2019-04-02
Mark crash aggregates dataset as obsolete

wlach:


wlach:
Let's just remove the reference altogether.

harterrt:
Looks like we missed two references: https://github.com/mozilla/firefox-data-docs/issues/267 .

PR #263, opened 2019-04-03
Bug 1539775 Clarify dates for DAU and MAU calculations

jklukas:


jklukas:
Thanks for the review, Jesse. I've simplified the text a bit and also copied it with slight alteration for the adau cookbook.

PR #268, opened 2019-04-18
Fix crash aggregates doc link

relud:


wlach:
Oops, sorry about this! Thanks for catching the problem @relud.

PR #271, opened 2019-05-02
Add note about using Colaboratory for accessing BigQuery

akkomar:
Colab might be convenient for some tasks, it's worth documenting that we can easily access BigQuery from there.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -42,6 +42,7 @@ CEP
 changesets
 chromehangs
 CircleCI
+Colaboratory
 config
 CommonMark
 CPOW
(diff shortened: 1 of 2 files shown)


akkomar:
Also pending https://bugzilla.mozilla.org/show_bug.cgi?id=1548495

akkomar:
@jmccrosky since you've been using Colab: this PR contains a link to example [notebook](https://colab.research.google.com/drive/1NCkpZBx3cRc08-g_JKMBrf5cIxAvSNMv) - would you mind taking a quick look and letting me know if it looks like a good starting point?

jmccrosky:
Looks great!  There are two things I might consider adding in the future:

I have a JSON service account key and in theory I should be able to use it so I don't have to manually authenticate every time I run the notebook?  I'd love to know how.

Maybe an example of writing to BQ?  I've been using pandas_gbq, although discovered it can't write DATE types (it does TIMESTAMP instead), so you might prefer to use a different approach as the example.  If you want, what I'm doing is: pandas_gbq.to_gbq(data, 'analysis.jmccrosky_test', project_id="moz-fx-data-derived-datasets", if_exists='append') for example.

PR #275, opened 2019-05-18
Add some historic events to analysis gotchas

mreid-moz:
Specifically this is to make note of [Bug 1550787](https://bugzilla.mozilla.org/show_bug.cgi?id=1550787) for future reference, but there are likely other noteworthy things to add here as well.

### DIFF
--- a/src/concepts/analysis_gotchas.md
+++ b/src/concepts/analysis_gotchas.md
@@ -8,7 +8,7 @@ This document is not about those traps. Instead, it is about quirks and pitfalls
 
 When looking at trends, it is helpful to be aware of events from the past that might impact comparisons with history. Here are a few to keep in mind:
 
-- **May 4 - May 11 2019** - [Telemetry source data deleted](https://blog.mozilla.org/blog/2019/05/09/what-we-do-when-things-go-wrong/). No source data is available this period.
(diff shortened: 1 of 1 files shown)


src/concepts/analysis_gotchas.md
```diff
@@ -4,6 +4,14 @@ When performing analysis on any data there are some mistakes that are easy to ma
 
 This document is not about those traps. Instead, it is about quirks and pitfalls specific to [Telemetry](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/index.html).
 
+### Notable historic events
+
+When looking at trends, it is helpful to be aware of events from the past that might impact comparisons with history. Here are a few to keep in mind:
+
+- **May 4 - May 11 2019** - [Telemetry source data deleted](https://blog.mozilla.org/blog/2019/05/09/what-we-do-when-things-go-wrong/). No source data is available this period.
```
jklukas:
```suggestion
- **May 4 - May 11 2019** - [Telemetry source data deleted](https://blog.mozilla.org/blog/2019/05/09/what-we-do-when-things-go-wrong/). No source data is available for this period and derived tables may have missing days or imputed values. Derived tables that depend on multiple days may have have affected dates beyond the deletion region.
```

PR #290, opened 2019-07-03
Add map access UDF to BigQuery documentation

sunahsuh:
Jason added the projects, and Jeff added the persistent UDF this morning so we can document it now!

cc @jasonthomas who I can't tag on the review for whatever reason

### DIFF
--- a/src/cookbooks/bigquery.md
+++ b/src/cookbooks/bigquery.md
@@ -117,6 +117,7 @@ namely that you must make use of the date partition fields for large tables (lik
 - Please read [_Query Optimizations_](bigquery.md#query-optimizations) section that contains advice on how to reduce cost and improve query performance.
 - re:dash BigQuery data sources will have a 10 TB data scanned limit per query. Please let us know in `#fx-metrics` on Slack if you run into issues!
 - There are no other partitioning fields in BigQuery versions of parquet datasets (e.g. `sample_id` is no longer a partitioning field and will not necessarily reduce data scanned).
+- There is no native map support in BigQuery. Instead, we are using structs with fields [key, value]. We have provided convenience functions to access these like key-value maps (described [below](bigquery.md#accessing-map-like-fields).)
 
 ### Projects with BigQuery datasets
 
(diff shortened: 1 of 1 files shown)


src/cookbooks/bigquery.md
```diff
@@ -190,6 +191,19 @@ You can create views in BigQuery if you have access via [GCP BigQuery Console](b
     - Prefix your view with your username. If your username is `username@mozilla.com` create a table with `username_my_view`.
 - See [Creating Views](https://cloud.google.com/bigquery/docs/views) documentation for detailed steps.
 
+### Accessing map-like fields
+
+BigQuery currently lacks native map support and our workaround is to use a STRUCT type with fields named [key, value]. We've created a user-defined function (UDFs) that provides key-based access with the signature: `udf.get_key(<struct>, <key>)`. The example below generates a count per `reason` key in the `event_map_values` field in the telemetry events table for Normandy unenrollment events from yesterday.
```
jklukas: There's a bit more nuance needed here, unfortunately. The `get_key` function is specific to the way we transform Parquet datasets to BigQuery, where we transform map types to a single-field struct that contains a repeated struct named `key_value` that contains key/value pairs. The `get_key` function expects a struct with a field literally named `key_value`. This works for datasets imported from Parquet, but not for maps unrolled by the GCP pipeline.

We should either include that nuance here or refactor get_key to be more general.

jklukas: See https://github.com/mozilla/bigquery-etl/pull/216 for the proposed refactor. It looks like we have no existing uses in bigquery-etl, so this change won't break any existing use cases.

jklukas: @relud is making an effort to get this resolved in the near future, so probably not worth mentioning for the time being.

Let's go ahead and merge the docs as-is, and then we will hopefully have a backend refactor in place by EOW that maintains the same interface described here.

src/cookbooks/bigquery.md
```diff
@@ -190,6 +191,19 @@ You can create views in BigQuery if you have access via [GCP BigQuery Console](b
...
+```sql
+SELECT udf.get_key(event_map_values, 'reason') AS reason,
+       COUNT(*) AS EVENTS
+FROM telemetry.events
+WHERE submission_date_s3 = DATE_SUB(CURRENT_DATE(), INTERVAL 1 DAY)
```
jklukas:
```suggestion
WHERE submission_date = DATE_SUB(CURRENT_DATE(), INTERVAL 1 DAY)
```

Both submission_date and submission_date_s3 exist now in the user-facing views. Probably best to encourage submission_date now for future compatibility.

PR #291, opened 2019-07-05
Add a style recommendation about explicit vs. implicit syntax.

mreid-moz:
Per discussion [here](https://github.com/mozilla/bigquery-etl/pull/210/files#r299609150).

cc @jklukas

### DIFF
--- a/src/concepts/sql_style.md
+++ b/src/concepts/sql_style.md
@@ -34,9 +34,13 @@ Always use uppercase for reserved keywords like `SELECT`, `WHERE`, or `AS`.
 1. Names must begin with a letter and may not end in an underscore.
 1. Only use letters, numbers, and underscores in variable names.
 
-## Aliasing
+## Be Explicit
 
-Always include the `AS` keyword when aliasing a variable,
+When choosing between explicit or implicit syntax, prefer explicit.
+
+### Aliasing
+
+Always include the `AS` keyword when aliasing a variable or table name,
 it's easier to read when explicit.
 
 **Good**
@@ -57,6 +61,45 @@ FROM
   main_summary
 LIMIT
   10
+```
+
+### Joins
+
+Always include the `JOIN` type rather than relying on the default join.
+
+**Good**
+```sql
+SELECT
+  submission_date,
+  e.key AS experiment_id,
+  e.value AS experiment_branch,
+  count(*) AS count
+FROM
+  telemetry.clients_daily
+CROSS JOIN
+  UNNEST(experiments) AS e(key, value)
+WHERE
+  submission_date > '20190701'
+  AND sample_id = '10'
+GROUP BY
+  1, 2, 3
+```
+
+**Bad**
+```sql
+SELECT
+  submission_date,
+  e.key AS experiment_id,
+  e.value AS experiment_branch,
+  count(*) AS count
+FROM
+  telemetry.clients_daily,
+  UNNEST(experiments) AS e(key, value)
+WHERE
+  submission_date > '20190701'
+  AND sample_id = '10'
+GROUP BY
+  1, 2, 3
 ```
 
 ## Left Align Root Keywords
--- a/src/concepts/sql_style.md
+++ b/src/concepts/sql_style.md
@@ -90,7 +90,7 @@ GROUP BY
 
 **Bad**
 ```sql
--- BigQuery SQL Syntax
+-- BigQuery Standard SQL Syntax
 SELECT
   submission_date,
   experiment.key AS experiment_id,
--- a/src/concepts/sql_style.md
+++ b/src/concepts/sql_style.md
@@ -28,7 +28,7 @@ Always use uppercase for reserved keywords like `SELECT`, `WHERE`, or `AS`.
 1. Use consistent and descriptive identifiers and names.
 1. Use lower case names with underscores, such as `first_name`.
    Do not use CamelCase.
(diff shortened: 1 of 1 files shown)


src/concepts/sql_style.md
```diff
@@ -59,6 +63,45 @@ LIMIT
...
+```sql
+SELECT
+  submission_date,
+  e.key AS experiment_id,
+  e.value AS experiment_branch,
+  count(*) AS count
+FROM
+  telemetry.clients_daily
+CROSS JOIN
+  UNNEST(experiments) AS e(key, value)
```
jklukas:
```suggestion
  UNNEST(experiments) AS experiment
```

I've seen several elegant unnest queries where the alias here is the non-plural version of the field name. I'd prefer `experiment` to `e` to encourage that style here, and perhaps even add a note encouraging that style, since UNNEST is going to be a common operation in the new BQ world.

I'm surprised to see the `e(key, value)` syntax. Is that actually renaming the fields in the struct?

src/concepts/sql_style.md
```diff
@@ -59,6 +63,45 @@ LIMIT
...
+WHERE
+  submission_date > '20190701'
+  AND sample_id = '10'
+GROUP BY
+  1, 2, 3
```
fbertsch: *ahem* *ahem* _explicit_ *ahem* *ahem* ;)

src/concepts/sql_style.md
fbertsch: This is the way to do it on Presto and Athena. We should begin marking the bq SQL statements. @mreid-moz I do recommend using just the `experiment` and marking this as a BQ-query. The alternative is keeping as-is and marking as a Presto query, then we can update everything at once.

mreid-moz: Yep, this is Presto syntax, and you can rename fields when `UNNEST`ing a map to do things like this:

```sql
SELECT
  submission_date_s3,
  experiment.experiment_id,
  experiment.experiment_branch,
  count(*) AS count
FROM
  telemetry.clients_daily
CROSS JOIN
  UNNEST(experiments) AS experiment(experiment_id, experiment_branch)
WHERE
  submission_date_s3 > '20190703'
  AND sample_id = '10'
GROUP BY
  1, 2, 3
```

+1 for the `experiment` naming, and +1 for including the SQL "flavour", I will update.

src/concepts/sql_style.md
mreid-moz: I *knew* this would come up :)

In Presto SQL, you cannot refer to the aliased column name in your `GROUP BY`, you must repeat the column expression (which can get ugly if it's more complicated than this example). In that case, I prefer the `GROUP BY 1` syntax over `GROUP BY experiment.key`. Particularly when it's `GROUP BY date_trunc(date_parse(...time string...), DAY)`

I just found out, though, that BigQuery *does* let you `GROUP BY` the aliased column name! Hooray!

fbertsch: That is great news!! I propose we make that the default in group-by expressions for BQ.

mreid-moz:
RFAL - I added a section about `GROUP BY 1, 2` - if that is too much in the weeds I can remove it.

chutten:
Thanks for the Cc.

A quick clarification: I'm not a fan of the numeric `GROUP BY`, I use it on re:dash because I have no choice (okay, using the full expression is _possible_, but it's not a choice). 

Thus: if using the aliases is possible, I'm all for calling it as preferred style. Can't wait.

Does Big Query allow use of resultset column aliases in `GROUPING SETS`? I ask because Presto/Athena don't even permit _expressions_ in `GROUPING SETS` (leading to me using an extra layer of CTE just to reify the column aliases as column names :( ).

fbertsch:
I have some bad news: [there is no `GROUPING SETS` in BQ.](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#syntax_3). Only rollup, which is _not_ equivalent. Yes, it sucks.

src/concepts/sql_style.md
```diff
@@ -59,6 +63,101 @@ LIMIT
   10
 ```
 
+### Joins
+
+Always include the `JOIN` type rather than relying on the default join.
+
+**Good**
+```sql
+-- BigQuery SQL Syntax
```
jklukas:
```suggestion
-- BigQuery Standard SQL Syntax
```

src/concepts/sql_style.md
```diff
@@ -59,6 +63,101 @@ LIMIT
...
+  AND sample_id = '10'
+GROUP BY
+  submission_date,
+  experiment_id,
+  experiment_branch
+```
+
+**Bad**
+```sql
+-- BigQuery SQL Syntax
```
jklukas:
```suggestion
-- BigQuery Standard SQL Syntax
```

src/concepts/sql_style.md
```diff
@@ -59,6 +63,101 @@ LIMIT
...
+-- Presto SQL Syntax
+SELECT
+  submission_date,
+  normalized_channel IN ('nightly', 'aurora', 'beta') AS is_prerelease,
+  count(*) AS count
+FROM
+  telemetry.clients_daily
+WHERE
+  submission_date > '20190701'
+GROUP BY 1, 2 -- Implicit grouping avoids repeating expressions
```
jklukas:
```suggestion
GROUP BY 
  1, 2 -- Implicit grouping avoids repeating expressions
```

PR #293, opened 2019-07-10
Document BigQuery Storage API Spark connector usage

akkomar:


akkomar:
> Looks good to me! I added a comment to the linked notebook about project names, otherwise +1

The project mentioned there is used to run queries (and billing), so it's not the one where the data is located. I have made it more clear in the description text.

PR #295, opened 2019-07-16
Include file through new (obsolete) path

badboy:
I'm not sure this should even be included there (given that it's in a directory called "obsolete").
It's still listed under "Obsolete Datasets"


wlach:
Thanks for noticing this, we should actually take this out and replace it with error aggregates. I'll do up a new PR for that in a bit!

PR #296, opened 2019-07-16
Use latest available mdbook & extension releases

badboy:
mdbook got several updates since we based this book on a fork.
It's now possible to use external preprocessors and both preprocessors
we use are now available as binary releases managed by me.


badboy:
Ah, I actually need to update the docs about this

badboy:
and that's fixed as well

PR #298, opened 2019-07-19
Update query optimization docs to reflect clustering

jklukas:


src/cookbooks/bigquery.md
```diff
@@ -224,7 +225,10 @@ To improve query performance and minimize the cost associated with using BigQuer
 
 - Avoid `SELECT *` by selecting only the columns you need
     - Using `SELECT *` is the most expensive way to query data. When you use `SELECT *` _BigQuery does a full scan of every column in the table._
-    - Applying a `LIMIT` clause to a `SELECT *` query does not affect the amount of data read. You are billed for reading all bytes in the entire table.
+    - Applying a `LIMIT` clause to a `SELECT *` query might not affect the amount of data read, depending on the table structure.
+      - Many of our tables are configured to use _clustering_ in which case a `LIMIT` clause does effectively limit the amount of data that needs to be scanned.
+      - Tables that include a `sample_id` field will usually have that as one of the clustering fields and you can efficiently scan random samples of users by specifying `WHERE sample_id = 0` (1% sample), `WHERE sample_id < 10` (10% sample), etc.
+      - To check whether your `LIMIT` and `WHERE` clauses are actually improving performance, you should see a lower value reported for actual "Data Scanned" by a query compared to the prediction ("This query will process X bytes") in STMO or the BigQuery UI.
```
harterrt: Fly by comment: 

If possible, it would be useful to have an order-of-magnitude guess at how bug this value should be.

Is 20 gigs too much for a one-off query? 200g?

src/cookbooks/bigquery.md
```diff
@@ -240,5 +244,6 @@ To improve query performance and minimize the cost associated with using BigQuer
...
     - See [approximate aggregation functions](https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#approximate-aggregate-functions) in the standard SQL reference.
+- Reference the data size prediction ("This query will process X bytes") in STMO and the BigQuery UI to help gauge the efficiency of your queries. You should see this number go down as you limit the range of `submission_date`s or include fewer fields in your `SELECT` statement. For clustered tables, this estimate won't take into account benefits from `LIMIT`s and `WHERE` clauses on clustering fields, so you'll need to compare to the actual "Data Scanned" after the query is run. [Queries are charged by data scanned at $5/TB](https://cloud.google.com/bigquery/pricing#on_demand_pricing) so each 200 GB of data scanned will cost $1; it can be useful to keep the data estimate below 200 GB while developing and testing a query to limit cost and query time, then open up to the full range of data you need when you have confidence in the results.
```
jklukas: @harterrt - How does this strike you?

harterrt: Awesome! Thank you!

src/cookbooks/bigquery.md
relud: perhaps it would be useful to point out how much a query costs ($5/TB) rather than how much is a reasonable amount scanned?

jklukas: Below, I do mention the $5/TB cost, but also provide 200 GB as a reasonable max size when developing a query. I do think that users want to have some concrete guidance about what "big" is.

relud: yeah, i hadn't gotten there yet >_<

PR #304, opened 2019-08-13
Bug 1571565 - Decommission retention_v1 dataset

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1571565)


src/datasets/obsolete/retention/reference.md
```diff
@@ -1,5 +1,9 @@
 # 1 Day Retention
 
+***As of 2019-08-13, this dataset has been deprecated and is no longer
```
fbertsch: How might someone calculate what was previously in the retention dataset?

acmiyaguchi: The process is involved -- it requires computing a HLL sketch per client dimension for each subsession date. The cut-off period is 10 days, which is enough time for 99% of pings the subsession date to arrive. The new-profile and main pings are merged together to form the source.

In BigQuery, this would involve cleaning the datasets and aligning the clients by a particular date, then using the `HLL_COUNT` udfs. Overall, there were only a few users, and I wouldn't recommend emulating this route due to the complexity.

fbertsch: I think it is a good idea to have a simple retention query, ala https://github.com/mozilla/stmo_core_product_metrics/blob/master/fennec/android/retention_day.sql

(If that seems sufficient, feel free to point people there).

acmiyaguchi:
@fbertsch This should be ready for another look now.

fbertsch:
Looks like the link-checker failed:
```
FILE: src/datasets/obsolete/retention/reference.md

[✖] https://github.com/mozilla/stmo_core_product_metrics/blob/master/fennec/android/retention_day.sql → Status: 404

[✖] https://github.com/mozilla/python_mozetl/blob/master/mozetl/engagement/retention/job.py → Status: 404
```

acmiyaguchi:
It looks like the first one is because stmo_core_product_metrics is a private repo, so I've removed it. Do you know of any other simple examples that would be accessible to the link checker?

fbertsch:
Hmm. I don't know of any others. Can we disable the link checker for that example?

acmiyaguchi:
Yep, that looks doable. I've added the repo to the ignored pattern list.

PR #308, opened 2019-08-19
Remove old Parquet docs

tdsmith:
We're going to stop writing new data to Parquet soon. Even in the current universe, these links mostly amount to "you can use `spark.read.parquet`", which isn't essential, so we can just remove it to avoid the pain of updating it.


src/cookbooks/parquet.md
```diff
@@ -1,65 +0,0 @@
...
-or in the reference documentation.
-
-# Writing Parquet Tables
-
-Saving a table to parquet is a great way to share an intermediate dataset.
-
-## Where to save data
-
-You can save data to a subdirectory of the following bucket:
-`s3://net-mozaws-prod-us-west-2-pipeline-analysis/<username>/`
```
mreid-moz: This one line is still of some use while we're on AWS. Can you find somewhere else to put this info temporarily? Otherwise, 💯 !

mreid-moz: `src/tools/spark.md` might be a good place to add this for now.

tdsmith:
rebased and r4r; snuck in a couple instances of s/ATMO/Spark/ while I was there.

PR #311, opened 2019-08-19
Remove longitudinal examples

tdsmith:
Longitudinal is EOL.


tdsmith:
The justification for removing these before removing the rest of the Longitudinal docs is that they're written in Presto SQL anyway and someone would have to port them. We wouldn't recommend that anyone do a new analysis with longitudinal today.

mreid-moz:
I'm holding off on this one for a bit - we're revisiting the "kill longitudinal" decision this week. If it lives on we'll probably want to port these examples to BQ. Longitudinal may still go away this quarter, in which case we can merge this as-is. Thanks!

PR #320, opened 2019-09-06
Organize cookbook section of TOC

tdsmith:
A cookbook should tell you how to do something, probably in a step-by-step way.

It also makes sense to separate cookbooks concerned with sending telemetry vs cookbooks concerned with accessing or interpreting telemetry.


tdsmith:
This deepens the yawning chasm between the organization of the TOC and the physical file structure. On the one hand, this seems bad for contributors; on the other hand, [cool URIs don't change](https://www.w3.org/Provider/Style/URI), so, idk, yolo.

PR #325, opened 2019-09-06
Active profiles

wcbeard:
This page documents the active_profiles table, and describes the model behind it

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -37,6 +37,8 @@ backend
 backends
 backfilled
 behaviour
+bgbb_airflow
+bgbb_lib
 Bhattacharyya
 BigQuery
 blocklist
@@ -179,6 +181,7 @@ prefs
 prepended
 protobuf
 pushlog
+PyPI
 PySpark
 RDBMS
 RDD
--- a/.spelling
+++ b/.spelling
@@ -37,8 +37,6 @@ backend
 backends
 backfilled
 behaviour
-bgbb_airflow
-bgbb_lib
 Bhattacharyya
 BigQuery
 blocklist
--- a/src/datasets/active_profiles.md
+++ b/src/datasets/active_profiles.md
@@ -1,0 +1,137 @@
+# Intro
+
+The `active_profiles` dataset gives client-level estimates of whether a profile
+is still an active user of the browser at a given point in time, as well as probabilistic forecasts
+of the client's future activity. These quantities are estimated by a model that attempts to infer
+and decouple a client's latent propensity to leave Firefox and become inactive, as well as their
+latent propensity to use the browser while still active.
+
+
+# Model
+The model generates predictions for each client by looking at just the recency and frequency of a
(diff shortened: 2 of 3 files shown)


.spelling
```diff
@@ -37,6 +37,8 @@ backend
 backends
 backfilled
 behaviour
+bgbb_airflow
+bgbb_lib
```
tdsmith: put these `` `in backticks` `` at the point of use instead

src/datasets/active_profiles.md
```diff
@@ -0,0 +1,140 @@
+# Intro
+
+The `active_profiles` dataset gives client-level estimates of whether a profile
+is still an active user of the browser at a given point in time, as well as probabilistic forecasts
```
ethompso28: specify "desktop" browser, and any particular selection criteria you had? (all OS and locations, for example?)

src/datasets/active_profiles.md
```diff
@@ -0,0 +1,140 @@
...
+
+# Model
+The model generates predictions for each client by looking at just the recency and frequency of a
+client's daily usage within the previous 90 day window. Usage is defined by the daily level binary
+indicator of whether they show up in `clients_daily` on a given day.
+
+The table contains columns related to these quantities:
+
+- `submission_date`: day marking the end of the 90 day window
+- `min_day`: first day in the window that the client was seen
```
ethompso28: is it possible then that the min_day could be 2 days before the 90 day window closes? or do you select min day to be at least 90 days ago (so that there was a chance to come back)? were there particular dates you looked at overall for this analysis?

src/datasets/active_profiles.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Attributes
+There are several columns that contain attributes of the client, like `os`, `locale`,
+`normalized_channel`, `normalized_os_version`, and `country`. `sample_id` is also included,
+which can be useful for quicker queries, as the table is clustered by this column in BigQuery.
+
+
+## Remarks on the model
+A way to think about the model that infers these quantities is to imagine a simple process
+where each client is given 2 coins when they become users, and that they flip each day.
```
ethompso28: are these fair coins, or are you weighting them somehow? it sounds like not, but want to check

ethompso28:
just had a couple questions on the model and a nit or two, but lgtm!

src/datasets/active_profiles.md
wcbeard: > is it possible then that the min_day could be 2 days before the 90 day window closes? 

It could be anywhere in the window, from the first to last day. I'll add a note explaining that.

> were there particular dates you looked at overall for this analysis?

The model is continuously running (so today it's looking at data in the past 90 days including June, July, August, and September). When I was developing the model I was looking at different date intervals, but don't remember which. Did you mean that or something else?

ethompso28: yeah, noting how far back the model goes might be good to add

PR #328, opened 2019-09-06
Explain feature gates

tdsmith:
Most experimentation is still pref-driven. Feature gates are a helpful wrapper
around prefs which are useful for a subset of experiments.


tdsmith:
@rhelmer This is just a quick take on what we'd discussed in #ask-experimenter; does this make sense?

rhelmer:
lgtm, thanks!

PR #333, opened 2019-09-19
active_profiles is release only for now

wcbeard:


wcbeard:
sorry @tdsmith, review required...

PR #336, opened 2019-09-27
Update BQ Storage API note

akkomar:
Clarify Storage API note. Going forward this should be extended with some guide on how-to-find-your-table.


src/cookbooks/bigquery.md
```diff
@@ -226,7 +226,7 @@ The table and view types referenced above are defined as follows:
...
 
-Spark and other applications relying on the BigQuery Storage API for data access may need to reference derived tables or historical ping tables directly rather than user-facing views in some cases, but we generally recommend instead that users run a query on top of user-facing views with the output saved in a destination table, which can then be accessed from Spark.
+Spark and other applications relying on the BigQuery Storage API for data access need to reference derived tables or historical ping tables directly rather than user-facing views. Unless the query result is relatively small we generally recommend instead that users run a query on top of user-facing views with the output saved in a destination table, which can then be accessed from Spark.
```
jklukas:
```suggestion
Spark and other applications relying on the BigQuery Storage API for data access need to reference derived tables or historical ping tables directly rather than user-facing views. Unless the query result is relatively large, we  recommend instead that users run a query on top of user-facing views with the output saved in a destination table, which can then be accessed from Spark.
```

PR #338, opened 2019-10-10
 Add cookbook for scheduling bigquery-etl queries with airflow 

Ben-Wu:
I think this would be useful because it wasn't immediately clear to me how to accomplish this use case.

### DIFF
--- a/src/cookbooks/bigquery-airflow.md
+++ b/src/cookbooks/bigquery-airflow.md
@@ -1,0 +1,107 @@
+# Scheduling BigQuery Queries in Airflow
+
+Queries in [BigQuery-ETL](https://github.com/mozilla/bigquery-etl) can be scheduled in [Airflow](https://github.com/mozilla/telemetry-airflow) to be run regularly with the results written to a table.
+
+<!-- toc -->
+
+## In BigQuery-ETL
+
+In the BigQuery-ETL project, queries are written in `/templates`.  The directory structure is based on the destination table: `/templates/{dataset_id}/{table_name}`.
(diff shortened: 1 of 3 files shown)


src/cookbooks/bigquery-airflow.md
```diff
@@ -0,0 +1,76 @@
+# Scheduling BigQuery Queries in Airflow
+
+Queries in [BigQuery-ETL](https://github.com/mozilla/bigquery-etl) can be scheduled in
```
jklukas: I'd prefer to consistently style this as `bigquery-etl` rather than `BigQuery-ETL` to match usage elsewhere.

src/cookbooks/bigquery-airflow.md
```diff
@@ -0,0 +1,76 @@
...
+    telemetry_derived.main_summary_v4
+WHERE
+  submission_date = @submission_date
+```
+
+`@submission_date` is a parameter that will be filled in by Airflow.
+
+After `/templates/example/client_ids/query.sql` is created, 
+`/script/generate_sql` can be run to generate the associated query in `/sql/examples/client_ids/query.sql`
+which is the query that will be run by the Airflow task.
```
jklukas: The details about directory structure and having to write queries under `templates/` are potentially going to change in the fairly near future, so I worry about codifying that here and neglecting to update these docs when we update the mechanics of bigquery-etl.

But, what you've written here is indeed a helpful, holistic look at the process. It would be weaker without these details. So I think I'm inclined to keep this as you've written it, but we need to be careful to remember this section exists when we update bigquery-etl.

PR #340, opened 2019-10-23
Bug 1585539 - add note about client_count decommission

akkomar:


jklukas:
> @akkomar we should probably get people to query `clients_daily` instead, just so they have a place to go.

The linked bug contains good documentation of the alternatives, which I think meets this need.

PR #347, opened 2019-11-06
Add another gotcha

mreid-moz:


wlach:
> Fine by me if theone is okay with it being public.

Note the link is Mozilla-only, and will fail the link-checker unless #351 is merged

PR #351, opened 2019-11-06
Don't check google docs links

wlach:
It seems like link checks against google docs are timing out. In general it's probably
not great practice to link against these kind of ephemeral resources, but in the
meanwhile this fixes builds.


wlach:
@mreid-moz does this seem sane?

wlach:
This is the link that's currently failing: https://docs.google.com/document/d/1r1PDQnqhsrPkft0pB46v9uhXGxR_FzK4laKJLGttXdA/edit -- it seems to have reverted from being world-readable to Mozilla only?

PR #354, opened 2019-11-12
bug 1564102 Create documentation for bmodata

emceeaich:


src/datasets/other/bmobugs/reference.md
```diff
@@ -0,0 +1,133 @@
+# bmobugs
+
+This dataset mirrors bugs (defects, enhancements, tasks) reported in the 
+[`Bugzilla`](https://bugzilla.mozilla.org/mozilla-central/) bug tracker.
+
+# Data Reference
+
+The dataset is accessible via [`STMO`](https://sql.telemetry.mozilla.org).
+Use the `????` table with the `????` data source.
```
mreid-moz:
```suggestion
Use the `eng_workflow.bmobugs` table with the `Telemetry (BigQuery)` data source.
```

src/datasets/other/bmobugs/reference.md
```diff
@@ -0,0 +1,133 @@
+# bmobugs
+
+This dataset mirrors bugs (defects, enhancements, tasks) reported in the 
+[`Bugzilla`](https://bugzilla.mozilla.org/mozilla-central/) bug tracker.
```
mreid-moz:
```suggestion
[`Bugzilla`](https://bugzilla.mozilla.org/) bug tracker.
```

src/datasets/other/bmobugs/reference.md
```diff
@@ -0,0 +1,133 @@
+# bmobugs
```
mreid-moz:
```suggestion
# `bmobugs`
```

src/datasets/other/bmobugs/reference.md
```diff
@@ -0,0 +1,133 @@
...
+in the past 7 days:
+
+```sql
+select
+    count(1)
+from
+    ???
+where
+    product = 'Core'
+    and date_diff('day', from_unixtime(creation_ts), now()) < 7
```
mreid-moz: 2nd attempt to add this comment :)

```
SELECT
  count(distinct bug_id)
FROM
  eng_workflow.bmobugs
WHERE
  product = 'Core'
  AND date_diff(current_date(), date(parse_timestamp('%Y-%m-%d %H:%M:%S', creation_ts)), DAY) <= 7
  AND date(submission_timestamp) >= '2019-01-01' -- required submission date filter
```

I added a `distinct` to the count because, as I understand it, the dataset might include multiple revisions of a given bug id.

I also added a `submission_timestamp` filter, which is required to minimize the amount of data scanned. It's not overly important for small datasets like `bmobugs`, but queries won't run without this implementation detail.

I ran the query itself, and it returned zero as the count - the most recent `submission_timestamp` in the data is November 9th, and the most recent `creation_ts` is November 2nd. Depending on how up-to-date the data is expected to be, it might be better for the example to use the past 14 days (which shows results) or instead count bugs per day of `creation_ts`.

emceeaich:
I've applied the suggested changes. There are some additional fields which we should include in the extract, but we can update the adapter and documentation after this initial patch lands.

mreid-moz:
The CI fail is unrelated to this change, and should be fixed by #356. Merging as-is.

PR #360, opened 2019-11-21
Optout ping replaced with deletion request ping

wlach:


src/datasets/ping_intro.md
```diff
@@ -87,14 +87,15 @@ contains a Telemetry Environment, [Crash Annotations][crash_annotations], and
...
 
-### "optout" ping
+### "deletion request" ping
 
 In the event a user opts out of Telemetry, we send one final
-["optout" ping][optout_ping] to let us know. We try exactly once to send it,
-discarding the ping if sending fails. It contains only the
-[common ping data][common_ping_data] and an empty payload.
+["deletion-request" ping][deletion_request_ping] to let us know. We try
+exactly once to send it, discarding the ping if sending fails. It contains
```
chutten: We actually can try more than once. In the event that the sending fails the ping remains pending and will be retried until it succeeds (or until it is discarded due to age or other reasons. TelemetrySend is complicated). This change in behaviour is because, unlike the "optout" ping which served a selfish purpose (allowing us to count optouts), the "deletion-request" ping is a request for us to delete a client's data. We want to make every effort to deliver that message.

src/datasets/ping_intro.md
```diff
@@ -87,14 +87,15 @@ contains a Telemetry Environment, [Crash Annotations][crash_annotations], and
...
+only the [common ping data][common_ping_data] and an empty payload.
 
-It was introduced in Firefox 63.
+It was introduced in Firefox 70, replacing the ["optout" ping][optout_ping]
```
chutten: 72, actually.

wlach: https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/data/deletion-request-ping.html implies 70, is it wrong?

src/datasets/ping_intro.md
wlach: I see. I don't know how much detail we want to go in here, so I just deleted the "exactly once" part.

The motivation for me filing/landing this is that the build is currently broken (because the opt out ping link now points to a non-existent location in master). If you have ideas to clarify/expand on what's here, we could always do a followup!

src/datasets/ping_intro.md
chutten: It is! (I'll file a bug) (I filed it: https://bugzilla.mozilla.org/show_bug.cgi?id=1598447 )

wlach:
The build process is still giving link checker errors, but they are unrelated to the changes here so imma gonna merge

PR #365, opened 2019-11-26
Allow link checking to fail

wlach:
The link checker tends to break for reasons outside of firefox-data-docs control--
it shouldn't be up to every PR submitted to fix it.

This PR also seperates out link and spell checking into their own jobs,
to make it more clear what's broken (or not) about a PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,32 @@
-language: rust
-addons:
-  apt:
-    packages:
-      - parallel
-install:
-  - npm install markdown-spellcheck markdown-link-check -g
-  - curl -LSfs https://japaric.github.io/trust/install.sh | sh -s -- --git rust-lang/mdBook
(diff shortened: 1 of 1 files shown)


mreid-moz:
I think it's a desirable effect that broken links end up getting fixed regularly, but I have definitely been annoyed by a failing build due to unrelated link problems myself.

This seems like a friendlier way to deal with broken links so 👍 .

wlach:
> How is one supposed to deal with the fact that links _might_ break for valid reasons?

Well that's covered by this PR, in that broken links will no longer fail the build. :)

@mreid-moz and I discussed how to make sure we check up on things occasionally. Short term, I think I'm going to have slackbot ping the channel asking us to check travis-ci for this every Monday. Longer term, I think I can have travis generate a report that will save the trouble of a click.

PR #368, opened 2019-11-27
Update search docs with gcp changes

Ben-Wu:
🕵️‍♂️


wlach:
I would suggest rebasing this on top of master and making sure CI passes before merging

PR #372, opened 2019-12-05
Bug 1595801 - Document utility function for determine a view's underlying table

relud:


src/tools/spark.md
```diff
@@ -70,22 +70,43 @@ Example of using the Storage API from Databricks:
 dbutils.library.installPyPI("google-cloud-bigquery", "1.16.0")
 dbutils.library.restartPython()
 
-# Read one day of pings and select a subset of columns.
+from google.cloud import bigquery
+
+
+def get_table(view):
+    bq = bigquery.Client()
```
jklukas:
```suggestion
    """Helper for determining what table underlies a user-facing view, since the Storage API can't read views."""
    bq = bigquery.Client()
```

relud:
> Can you also make sure this gets copied into the example Databricks notebook for accessing BQ?

sure, where is that?

jklukas:
It's linked from https://docs.telemetry.mozilla.org/cookbooks/bigquery.html#on-databricks

PR #380, opened 2019-12-13
Update retention documentation.

jmccrosky:
Getting the basic shape of things into place.  Please take a look and let me know if you have any suggestions or concerns.

### DIFF
--- a/src/cookbooks/retention.md
+++ b/src/cookbooks/retention.md
@@ -83,7 +83,7 @@ SELECT
   SAFE_DIVIDE(SUM(active_in_weeks_0_and_1), SUM(active_in_week_0)) AS one_week_retention
 FROM
   base
-WHERE submission_date = "2019-12-01"
+WHERE date = "2019-12-01"
 ```
 
 ## Confounding Factors
--- a/src/cookbooks/retention.md
+++ b/src/cookbooks/retention.md
@@ -8,21 +8,21 @@ Retention measures the proportion of users that are *continuing* to use Firefox,
 
 There is currently active research into retention metrics and we expect our conceptual and data model around retention to evolve in the coming months.  This page will be updated.  However, in the meantime, we want to be clear about our standard retention metrics for use right now.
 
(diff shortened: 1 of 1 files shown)


src/cookbooks/retention.md
```diff
@@ -1,225 +1,91 @@
...
+Among new profiles created on the day specified, what proportion (out of 1) meet the usage criterion during the week beginning one week after the day specified.
 
-```python
-{1: {'end': 13, 'start': 7},
- 2: {'end': 20, 'start': 14},
- 3: {'end': 27, 'start': 21},
- 4: {'end': 34, 'start': 28},
- 5: {'end': 41, 'start': 35},
- 6: {'end': 48, 'start': 42}}
+Note that we use a new profile definition that relies on the `profile_creation_date` and requires that a main ping be sent within one week of the `profile_creation_date`.  This differs from analysis using new profile pings, but allows valid comparison over time, which new profile pings do not due to the increase adoption of versions of the browser recent enough to send new profile pings.
```
jklukas:
```suggestion
Note that we use a new profile definition that relies on the `profile_creation_date` and requires that a main ping be sent within one week of the `profile_creation_date`.  This differs from analysis using new profile pings, but allows valid comparison over time. New profile pings do not allow comparison over time due to the increased adoption of versions of the browser recent enough to send new profile pings.
```

src/cookbooks/retention.md
```diff
@@ -1,225 +1,91 @@
...
+  AS (
+    BIT_COUNT(x >> (7 * n) & (0x7F)) > 0
+  );
+CREATE TEMP FUNCTION
+  udf_bitpos( bits INT64 ) AS ( CAST(SAFE.LOG(bits & -bits, 2) AS INT64));
+
+WITH base AS (
+  SELECT
+    client_id,
+    submission_date,
```
jklukas:
```suggestion
    DATE_SUB(submission_date, INTERVAL 13 DAY) AS date,
```

We should probably include the 13 day shift as part of this example query, and modify the final SELECT to reference `date = "2019-12-01".

jmccrosky:
Thanks you two for the approval and thanks Jeff for the great suggestions!

PR #383, opened 2019-12-16
Add documentation for static datasets

Ben-Wu:


src/datasets/static.md
```diff
@@ -0,0 +1,29 @@
+# Static Datasets
+
+Tables containing static data exist in the `static` dataset in BigQuery.
+These tables are generated from CSV's named `data.csv` in the `templates`
```
jklukas:
```suggestion
These tables are generated from CSV files named `data.csv` in subdirectories of the `templates/static` 
```

src/datasets/static.md
```diff
@@ -0,0 +1,29 @@
+# Static Datasets
+
+Tables containing static data exist in the `static` dataset in BigQuery.
+These tables are generated from CSV's named `data.csv` in the `templates` 
+directory in [`bigquery-etl`](https://github.com/mozilla/bigquery-etl/tree/master/templates/static).
+
+## Creating a Static Table
+
+Static tables can be created in any dataset in bigquery-etl.  To create a new table,
```
jklukas: You link specifically to temlates/static above. I think that in these docs we should encourage all static tables to reside under `static`. We can mention that the machinery allows them to live elsewhere, but but doing so should be carefully considered.

PR #389, opened 2020-01-02
1601005: Document how to access Glean data

mdboom:


src/concepts/glean/accessing_glean_data.md
```diff
@@ -0,0 +1,71 @@
...
+    - metrics
+```
+
+It would be availabe in the column `metrics.boolean.metrics_default_browser`.
+
+```sql
+-- Count number of pings where Fenix is the default browser
+SELECT
+  COUNT(*),
+  COUNTIF(metrics.metrics.boolean.metrics_default_browser)
```
mdboom: The requirement for the extra `metrics` here isn't great (since the name of the table matches the name of the record, SQL gets confused otherwise).  Any SQL tricks to avoid that?  If not, I'll go back and document this fiddly detail.

jklukas: You can avoid this is you give the target table a name (`FROM org_mozilla_fenix.metrics AS m`), which probably leads to a cleaner-looking query, but it's then strange that you've given something a name you never reference.

I do think that this is a fiddly detail that needs to be documented as it's going to continue biting people.

For the purposes of this example, though, my personal preference would be to do the rename trick, and add a comment about why you're doing so.

src/concepts/glean/accessing_glean_data.md
```diff
@@ -0,0 +1,71 @@
...
+FROM
+  org_mozilla_fenix.metrics
```
jklukas:
```suggestion
  -- We give the table an alias so that the table name `metrics` and field name `metrics` don't conflict.
  org_mozilla_fenix.metrics AS m
```

src/concepts/glean/accessing_glean_data.md
```diff
@@ -0,0 +1,71 @@
...
+This document describes how to access Glean data from an SQL query, such as in [Redash](https://sql.telemetry.mozilla.org).
+
+## Selecting the correct table
+
+Each ping type is recorded in its own table, and these tables are named using `{application_id}.{ping_type}`.
+For example, for Fenix, the application id is `org_mozilla_fenix`, so its `metrics` pings are available in the table `org_mozilla_fenix.metrics`.
+
+## Selecting columns
+
+Columns of data are named using dot notation to organize them into groups.
```
jklukas: This is perhaps an oversimplification, since at the BigQuery level these are nested structs and the dot notation is used to descend into them.

How about "Fields are nested inside BigQuery `STRUCT`s to organize them into groups, and we can use dot notation to specify individual subfields in a query."

src/concepts/glean/accessing_glean_data.md
```diff
@@ -0,0 +1,71 @@
...
+For example, columns containing Glean's built-in client information all begin with the `client_info.` prefix.
+
+The top-level groups are:
+
+* `client_info`: [Client information provided by Glean](https://mozilla.github.io/glean/book/user/pings/index.html#the-client_info-section).
+* `ping_info`: [Ping information provided by Glean](https://mozilla.github.io/glean/book/user/pings/index.html#the-ping_info-section).
```
jklukas: I am actively working on https://github.com/mozilla/gcp-ingestion/issues/633 right now which will make new `ping_info.parsed_start_time` and `ping_info.parsed_end_time` fields available in the user-facing views for glean pings. We can have a separate docs PR once that is change is merged and deployed, but this definitely seems like a good place to advertise the existence of those parsed fields.

src/concepts/glean/accessing_glean_data.md
mdboom: That's exactly why I asked you to review :)  Thanks.

src/concepts/glean/accessing_glean_data.md
mdboom: Agreed!  Let's add those here when that lands.

Also, this makes me realize there are other fields that are beyond the ping content that should probably be documented here, such as `submission_timestamp` and `document_id`.

src/concepts/glean/accessing_glean_data.md
```diff
@@ -0,0 +1,73 @@
...
+  org_mozilla_fenix.metrics AS m
+WHERE
+  date(submission_timestamp) = '2019-11-11'
+```
+
+### The `events` group
+
+Custom events in the `events` section have a different structure.
+
+Documentation TBD.
```
Dexterp37: Is there a bug filed for this? Can we mention it here?

src/concepts/glean/accessing_glean_data.md
```diff
@@ -0,0 +1,73 @@
+# Accessing Glean data
+
+This document describes how to access Glean data from an SQL query, such as in [Redash](https://sql.telemetry.mozilla.org).
+
+## Selecting the correct table
+
+Each ping type is recorded in its own table, and these tables are named using `{application_id}.{ping_type}`.
+For example, for Fenix, the application id is `org_mozilla_fenix`, so its `metrics` pings are available in the table `org_mozilla_fenix.metrics`.
```
Dexterp37:
```suggestion
For example, for Fenix, the application id is `org.mozilla.fenix`, so its [`metrics`](https://mozilla.github.io/glean/book/user/pings/metrics.html) pings are available in the table `org_mozilla_fenix.metrics`.
```

src/concepts/glean/accessing_glean_data.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+### The `metrics` group
+
+Custom metrics in the `metrics` section have two additional levels of indirection in their column name: they are organized by the metric type, and then by their category: `metrics.{metric_type}.{category}_{name}`. 
+
+For example, suppose you had the following `boolean` metric defined in a `metrics.yaml` file (abridged for clarity):
+
+```yaml
+metrics:
```
Dexterp37: nit: let's not use `metrics` as a category name, as we actively discourage this in the glinter.

Should we, instead, use: `browser` as a category and `is_default` as the metric name?

src/concepts/glean/accessing_glean_data.md
mdboom: [There is](https://bugzilla.mozilla.org/show_bug.cgi?id=1606836).  I'll link to it here.

PR #395, opened 2020-01-15
Migrate GUD documentation to DTMO.

jmccrosky:
Should be equivalent to existing documentation at https://docs.google.com/document/d/1sIHCCaJhtfxj-dnbInfuIjlMRhCFbEhFiBESaezIRwM/

### DIFF
--- a/src/tools/gud.md
+++ b/src/tools/gud.md
@@ -1,0 +1,170 @@
+# Mozilla Growth & Usage Dashboard
+#### *overall: jmccrosky@ | data: jklukas@ | frontend: hulmer@ (updated 2020-1-15)*
+___
+
+**The Mozilla Growth & Usage Dashboard is in general release and can be found [here](https://growth-stage.bespoke.nonprod.dataops.mozgcp.net/).**
+
+___
+
+# Overview
+
+The Mozilla Growth & Usage Dashboard (GUD) is a tool to visualize growth metrics in a standard way across Mozilla’s products.  It is intended to be quite intuitive to use, but we provide some brief documentation here.  Please reach out (`#gud` on slack) if anything is unclear or you are unsure how to use GUD for your purposes.
+
+# Quick Start
+
+To visualize a particular metric (such as `MAU` or `Retention`) for a particular product:
+
+1. By default, all metrics are shown.  To zoom into a particular metric, choose it from the `METRIC` selector on the left.
+
+2. Choose the product of interest from the `PRODUCT / USAGE CRITERIA` selector on the left.  Some products may have multiple possible usage criteria, corresponding to different types of use of the product.
+
+    * New profile creation is considered one of these usage criteria; thus, to determine the rate of new profile creation for a product, select that product’s “`Profile Created`” usage criteria and look at the `DAU` metric.
+
+You may also restrict to particular slices (for example, country) and to a particular date range using the additional selectors on the left.
+
+# Data Model
+
+The key to understanding how to use GUD is to understand the GUD data model.  The key concepts are `usage criteria`, `slicing dimensions`, and `metrics`.
+
+* A `usage criterion` defines what a metric is measuring.  It generally specifies a product and what specific use of that product qualifies as active.  For example, “Any Fenix Activity” measures activity for all Fenix profiles that send a telemetry pings.  Similarly, “Opened DevTools” means that we measure activity for all Firefox profiles that send a telemetry ping indicating that DevTools was opened.
+
+* A `dimension` allows slicing to a subset of profiles according to characteristics of those profiles.  Available dimensions include: country, channel, OS, etc.  A slice is a set of particular values within dimensions, for example, “country is US” is one slice and “country is either US or DE and channel is release” is another.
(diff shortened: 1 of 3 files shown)


src/tools/gud.md
```diff
@@ -0,0 +1,170 @@
...
+
+___
+
+# Overview
+
+The Mozilla Growth & Usage Dashboard (GUD) is a tool to visualize growth metrics in a standard way across Mozilla’s products.  It is intended to be quite intuitive to use, but we provide some brief documentation here.  Please reach out (`#gud` on slack) if anything is unclear or you are unsure how to use GUD for your purposes.
+
+# Quick Start
+
+To visualize a particular metric (such as `MAU` or `Retention`) for a particular product:
```
jklukas: Most of the terms in backticks in this article probably don't need them, or should be italicized with underscores instead. In particular, MAU and retention both seem like natural words that don't need any adornment.

src/tools/gud.md
```diff
@@ -0,0 +1,170 @@
...
+
+1. By default, all metrics are shown.  To zoom into a particular metric, choose it from the `METRIC` selector on the left.
+
+2. Choose the product of interest from the `PRODUCT / USAGE CRITERIA` selector on the left.  Some products may have multiple possible usage criteria, corresponding to different types of use of the product.
+
+    * New profile creation is considered one of these usage criteria; thus, to determine the rate of new profile creation for a product, select that product’s “`Profile Created`” usage criteria and look at the `DAU` metric.
```
jklukas:
```suggestion
    * New profile creation is considered one of these usage criteria; thus, to determine the rate of new profile creation for a product, select that product’s "`Profile Created`" usage criterion and look at the `DAU` metric.
```

src/tools/gud.md
```diff
@@ -0,0 +1,170 @@
...
+
+You may also restrict to particular slices (for example, country) and to a particular date range using the additional selectors on the left.
+
+# Data Model
```
jklukas: This explanation largely overlaps with existing docs for the Exact MAU tables: https://docs.telemetry.mozilla.org/datasets/bigquery/exact_mau/reference.html

We should probably consolidate so that we don't duplicate these explanations, and make sure to reference whatever the canonical location is from the other location.

src/tools/gud.md
```diff
@@ -0,0 +1,170 @@
...
+
+The country (as extracted through IPGeo) of the profile’s pings.
+
+### Channel
+
+The channel of the profile’s pings.
+
+# Bug reports / Feature Requests
+
+Please direct all feedback to ``#gud` on slack.
```
jklukas:
```suggestion
Please direct all feedback to `#gud` on slack.
```

src/tools/gud.md
```diff
@@ -0,0 +1,170 @@
...
+
+Whether the profile’s pings were sent from Mac OS, Windows, or Linux.
+
+### Language
+
+The language (as extracted from the locale) of the profile’s pings.
+
+### Country
+
+The country (as extracted through IPGeo) of the profile’s pings.
```
jklukas:
```suggestion
The country (as extracted through GeoIP lookup) of the profile’s pings.
```

src/tools/gud.md
```diff
@@ -0,0 +1,170 @@
...
+### Fennec
+
+#### Any Fennec Android Activity
+
+The profile has sent a telemetry ping from Fennec Android on the day in question.
+
+#### Any Fennec iOS Activity
+
+The profile has sent a telemetry ping from Fennec iOS on the day in question.
+### Focus
```
jklukas:
```suggestion

### Focus
```

src/tools/gud.md
jmccrosky: I see your point, but also kind of feel it really needs to be in both places.  I think it's pretty essential to really understanding how to use GUD, so want to have it here, but also see that it needs to be in the exact MAU table reference.  Does markdown support an include?  Maybe that's a partial solution to at least restrict that they're in sync...  I'm also open to being convinced that one could just link to the other, but don't see it that way now.

jklukas: I agree that it's pretty central in both cases. I'm okay with duplicating for now, but perhaps we should at least make sure we point out in both articles pretty early on that a very similar data model exists for the other case, so these are obviously cross-linked. I don't want to block this on figuring out a technical solution.

src/tools/gud.md
jmccrosky: My intent was to call out words that correspond to UI elements in the dashboard, like the PRODUCT / USAGE CRITERIA selector, or specific menu items like metrics and such.  Do you think this is not helpful?  Or helpful but better done with italics or something?

jklukas: I see your point. Let's roll with it.

src/tools/gud.md
jmccrosky: Sounds fair.  I made the change to better cross-link.  I'll wait until Friday morning to merge in case you want to review that, or any of the other reviewers want to weigh in.

PR #401, opened 2020-01-22
Add a historical reference section and reorganize TOC

wlach:
It's not really helpful to mix in things we don't support
with things we do. For clarity, let's move that stuff to its
own section at the very bottom.


wlach:
> Changing where articles exist in the summary doesn't affect their URLs, right? It would be nice to make sure any existing links to the old AWS pipeline docs, etc. will still work. I think they will.

Yup, everything is still in its old location. Just the top-level organization in the table of contents is different. Thanks for the fast review!

PR #413, opened 2020-02-19
Clean up "choosing a desktop dataset"

wlach:
This whole article needs more work, but removing documentation about obsolete datasets (which
are covered elsewhere in a reference) and irrelevant documentation about random mobile 
datasets are no brainers.


wlach:
Build failure fixed by #414

PR #424, opened 2020-03-09
Update links

wlach:
* Remove analysis.tmo *from* ignore list (we shouldn't reference that anywhere)
* Add some circle ci example links *to* ignore list

### DIFF
--- a/.linkcheck.json
+++ b/.linkcheck.json
@@ -9,7 +9,6 @@
         { "pattern": "^https://sql.telemetry.mozilla.org" },
         { "pattern": "^https://github.com/mozilla/stmo_core_product_metrics" },
         { "pattern": "^https://github.com/mozilla-services/cloudops-infra" },
-        { "pattern": "^https://analysis.telemetry.mozilla.org" },
         { "pattern": "^https://docs.google.com" }
     ]
 }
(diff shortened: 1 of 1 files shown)


wlach:
> Are we excluding the circleci link due to auth requirements?

That was my thought, yeah. Maybe we should call out that the links won't work in the cookbook? https://circleci.com/gh/mozilla/pensieve/edit doesn't work for me either

wlach:
This is the doc fwiw https://docs.telemetry.mozilla.org/cookbooks/deploying-containers.html

jklukas:
> > Are we excluding the circleci link due to auth requirements?
> 
> That was my thought, yeah. Maybe we should call out that the links won't work in the cookbook? https://circleci.com/gh/mozilla/pensieve/edit doesn't work for me either

It doesn't work for me either, which is surprising. I've double-checked that I'm authenticated via GH creds. We should ask @sunahsuh about permissions on that project. We shouldn't have the link at all if basically no one reading the docs will be able to follow it.

I think we're good to merge this PR though.

wlach:
Yeah, let's merge it. Not good to have the build broken. Filed #425 to investigate the issue with the doc.

PR #427, opened 2020-03-13
Add a page for standard metric definitions.

mreid-moz:
Create a place to build out definitions for standard metrics, as well
as improving discoverability by bringing it to the top of the
navigation and linking from the Getting Started page.

The definitions are based directly on GUD.


src/SUMMARY.md
```diff
@@ -1,14 +1,15 @@
 * [Firefox Data Documentation](introduction.md)
```
harterrt: Dropping the bullet makes the numbering in the side bar work better. 

```suggestion
[Firefox Data Documentation](introduction.md)
```

src/SUMMARY.md
```diff
@@ -1,14 +1,15 @@
...
+  * [Reporting a problem](concepts/reporting_a_problem.md)
   * [Gaining Access](concepts/gaining_access.md)
   * [Analysis Quick Start](concepts/analysis_intro.md)
   * [Choosing a Desktop Dataset](concepts/choosing_a_dataset.md)
   * [Choosing a Mobile Dataset](concepts/choosing_a_dataset_mobile.md)
   * [Intro to STMO](tools/stmo.md)
   * [Common Analysis Gotchas](concepts/analysis_gotchas.md)
   * [Optimizing Queries](concepts/sql_optimization.md)
+  * [Terminology](concepts/terminology.md)
   * [Getting Help](concepts/getting_help.md)
```
harterrt: A few suggestions on ordering. 

* Keep *Standard Metrics* as part of *Getting Started*. 
* Consider merging *Reporting a problem* into the *Meta* or *Getting Help* section
* General reordering for flow
 

```suggestion
* [Getting Started](concepts/getting_started.md)
  * [Gaining Access](concepts/gaining_access.md)
  * [Analysis Quick Start](concepts/analysis_intro.md)
  * [Standard Metrics](metrics/index.md)
  * [Common Analysis Gotchas](concepts/analysis_gotchas.md)
  * [Choosing a Desktop Dataset](concepts/choosing_a_dataset.md)
  * [Choosing a Mobile Dataset](concepts/choosing_a_dataset_mobile.md)
  * [Intro to STMO](tools/stmo.md)
  * [Optimizing Queries](concepts/sql_optimization.md)
  * [Terminology](concepts/terminology.md)
  * [Getting Help](concepts/getting_help.md)
  * [Reporting a problem](concepts/reporting_a_problem.md)
```

src/metrics/index.md
```diff
@@ -0,0 +1,53 @@
+# Standard Metrics
+
```
harterrt: I suggest adding an intro to help the reader get their bearings.

```suggestion
 This section provides an overview of standard metrics used at Mozilla. 
 Here you'll find the definitions and descriptions for each.
 
 For a deep dive into these metrics, see [the GUD documentation](../tools/gud.md).
```

src/metrics/index.md
```diff
@@ -0,0 +1,53 @@
...
+Among new profiles created on the day specified, what proportion (out of 1) are
+active during the week beginning one week after the day specified.
+
+### 1-Week Retention
+
+Among profiles that were active at least once in the week starting on the
+specified day, what proportion (out of 1) are active during the following week.
+
+## Frequently Asked Questions
```
harterrt: I'm concerned these are too prominent. Can we add them to the GUD article?

Since these are in *Getting Started* I feel like I need to understand why these questions are important for all new analysts to see.

jmccrosky:
Thanks Mark!  Looking at the discussion of where to put it, I kind of like having a top-level metrics section, which is more of a reference and can grow to the point that most people might not want to read through it in its entirety.  We can then have a shorter metrics summary in getting started that also links to that top-level section.

What do you both think?

harterrt:
> Thanks Mark! Looking at the discussion of where to put it, I kind of like having a top-level metrics section, which is more of a reference and can grow to the point that most people might not want to read through it in its entirety. We can then have a shorter metrics summary in getting started that also links to that top-level section.

I agree that we should have a top level *Metrics* session in addition to the short introduction in *Getting Started* that this PR introduces. I do not want to block this PR on the *Metrics* section though.

jmccrosky:
Sounds totally fair.  I'm happy for Mark to commit this and we can iterate from there.

src/metrics/index.md
wlach: It looks like there are also more detailed definitions later in the docs under "Telemetry Behaviour Reference", might be worth a mention of that as well.

src/metrics/index.md
mreid-moz: I would argue that these particular things are important enough to be read by everyone when getting started. They're not super obvious, and (other than the first one) have been subject to debate / change in the past.

mreid-moz:
Sorry it's taken me so long to come back to this. Thanks for the feedback all, this is RFAL.

PR #428, opened 2020-03-19
Add a first draft of segments documentation.

felixlawrence:
This hopefully provides a place to communicate the direction I think we should be heading, as well as discuss some of the more open questions.

See also the corresponding [bigquery-etl PR](https://github.com/mozilla/bigquery-etl/pull/825), and the [internal doc about the regular user v1 set of segments](https://docs.google.com/document/d/1eI7dXB6iYl_8gQ0wXdPiilLDe4bzAShdgigrDCqThvg/).

### DIFF
--- a/src/concepts/segments.md
+++ b/src/concepts/segments.md
@@ -11,14 +11,14 @@ For example, "In this experiment, how do new users react to this feature,
 and how does this differ from established users?". 
 Or "DAU moved dramatically - 
 is this restricted to users in this particular country (i.e. a segment) 
-where there's an event happening, 
+where there's an event happening, which would raise our suspicions,
 or is it global and therefore not solely due to that event?"
 
 ## Versioning
 
 We are building out our library of segments, 
 and we want room to iterate to improve them in the future. 
-So please quote segments' versions with their names, 
+So please quote segments' versions with their names, e.g. "regular users v1"
 so that your communication is forwards compatible.
 
 ## Current segments
--- a/src/concepts/segments.md
+++ b/src/concepts/segments.md
@@ -28,27 +28,20 @@ so that your communication is forwards compatible.
 This is a set of three segments. 
 On a given day, every client falls into exactly one of these segments.
 Each client's segment is stored in `telemetry.clients_last_seen.segment_regular_users_v1`.
-(FIXME: or it will be if the [PR](https://github.com/mozilla/bigquery-etl/pull/825) is accepted in a form similar to its current state)
+
 
 *Regular users v1* is defined as 
-clients who have at least two days where they browsed >=5 URIs, 
-in *all* of the last four weeks. 
+clients who browsed >=5 URIs on at least _one_ of the previous 6 days, and browsed >=5 URIs on _at least two_ days in each of the three weeks before that. 
 Observationally, on any given day this segment seems to contain a large fraction of our users 
 and has exceptionally high retention.
 
 *New/irregular users v1* is defined as 
-clients who browsed >=5 URIS on at least two days in *none* of the last four weeks. 
+clients who did not browse >=5 URIs on any of the previous 6 days, and browsed >=5 URIs on _less than two_ days in each of the three weeks before that.
 This is a smaller segment of our users and has low retention 
 (though "activation" is likely a more relevant word than "retention" for many of these clients).
 
-*Semi-regular users v1* is the other users: the segment is defined as 
-clients who browsed >=5 URIs on at least two days in *some but not all* of the last four weeks. 
+*Semi-regular users v1* contains the other clients.  
 This seems to be a smaller segment of our users, and has relatively high retention.
-
-In each of these definitions, 
-"the last four weeks" refer to 7 day periods that end on the day of week 
-before the first day for which we want to study the user. 
-They do not necessarily run Monday through Sunday.
 
 ## Obsolete segments
 
@@ -75,11 +68,10 @@ Segments are found as columns in the `clients_last_seen` dataset: the segment li
 ### WAU and MAU
 
 Users can change segments part way through a week or a month. 
-When computing MAU for a segment, 
-include all clients that fit that segment for at least one day they were active in the month, 
-regardless of whether they fit a contradictory segment on another active day. 
-This means that the sum of MAU for each non-overlapping segment can be larger than MAU.
-(FIXME: This point needs special attention in the review process - it appears to be the best option)
+At present, 
+we have not settled the questions of whether and how to split MAU by segment,
+and whether and how to measure MAU for each segment.
+So stick to DAU for now.
 
 
 ### Example queries
@@ -88,7 +80,7 @@ DAU for regular users v1:
 ```lang=sql
 SELECT
     submission_date,
-    COUNT(DISTINCT client_id) AS dau_regular_v1
+    COUNT(*) AS dau_regular_v1
 FROM moz-fx-data-shared-prod.telemetry.clients_last_seen
 WHERE
     submission_date BETWEEN '2020-01-01' AND '2020-03-01'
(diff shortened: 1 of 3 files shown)


src/concepts/segments.md
```diff
@@ -0,0 +1,134 @@
...
+Typically you'll use segments to gain more insight into what is going on, by asking 
+"Regarding the thing I'm interested in, 
+do users in different segments behave differently, 
+and what insights does this give me into the users and the product?" 
+For example, "In this experiment, how do new users react to this feature, 
+and how does this differ from established users?". 
+Or "DAU moved dramatically - 
+is this restricted to users in this particular country (i.e. a segment) 
+where there's an event happening, 
+or is it global and therefore not solely due to that event?"
```
jmccrosky: A little pedantic, but the phenomenon being restricted to a particular country doesn't necessarily tell us what caused the phenomenon.  I'd consider rephrasing to avoid promoting this kind of thinking.

src/concepts/segments.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Versioning
+
+We are building out our library of segments, 
+and we want room to iterate to improve them in the future. 
+So please quote segments' versions with their names,
```
jmccrosky: Maybe give an example?

src/concepts/segments.md
```diff
@@ -0,0 +1,134 @@
...
+### Regular users v1, Semi-regular users v1, New/irregular users v1
+
+This is a set of three segments. 
+On a given day, every client falls into exactly one of these segments.
+Each client's segment is stored in `telemetry.clients_last_seen.segment_regular_users_v1`.
+(FIXME: or it will be if the [PR](https://github.com/mozilla/bigquery-etl/pull/825) is accepted in a form similar to its current state)
+
+Regular users v1 is defined as 
+clients who have at least two days where they browsed >=5 URIs, 
+in *all four* of the last weeks.
```
jmccrosky: "in all of the last four weeks" might be clearer

src/concepts/segments.md
```diff
@@ -0,0 +1,134 @@
...
+## Current segments
+
+### Regular users v1, Semi-regular users v1, New/irregular users v1
+
+This is a set of three segments. 
+On a given day, every client falls into exactly one of these segments.
+Each client's segment is stored in `telemetry.clients_last_seen.segment_regular_users_v1`.
+(FIXME: or it will be if the [PR](https://github.com/mozilla/bigquery-etl/pull/825) is accepted in a form similar to its current state)
+
+Regular users v1 is defined as
```
jmccrosky: I'd put the segment names in a different text style.

src/concepts/segments.md
```diff
@@ -0,0 +1,134 @@
...
+(FIXME: or it will be if the [PR](https://github.com/mozilla/bigquery-etl/pull/825) is accepted in a form similar to its current state)
+
+Regular users v1 is defined as 
+clients who have at least two days where they browsed >=5 URIs, 
+in *all four* of the last weeks. 
+Observationally, on any given day this segment seems to contain a large fraction of our users 
+and has exceptionally high retention.
+
+Semi-regular users v1 is defined as 
+clients who browsed >=5 URIs on at least two days in each of *one, two or three* of the last four weeks.
```
jmccrosky: This line is confusing.

src/concepts/segments.md
```diff
@@ -0,0 +1,134 @@
...
+Segments are found as columns in the `clients_last_seen` dataset: the segment listed for a client on a `submission_date` is valid for that `submission_date` because it is computed only using behavioral data collected _before_ the `submission_date`.
+
+### WAU and MAU
+
+Users can change segments part way through a week or a month. 
+When computing MAU for a segment, 
+include all clients that fit that segment for at least one day they were active in the month, 
+regardless of whether they fit a contradictory segment on another active day. 
+This means that the sum of MAU for each non-overlapping segment can be larger than MAU.
+(FIXME: This point needs special attention in the review process - it appears to be the best option)
```
jmccrosky: I'd disagree.  I think that for any given metricXprofileXdate, there should be a single segment.  We can discuss elsewhere maybe.

src/concepts/segments.md
```diff
@@ -0,0 +1,134 @@
...
+In each of these definitions, 
+"the last four weeks" refer to 7 day periods that end on the day of week 
+before the first day for which we want to study the user. 
+They do not necessarily run Monday through Sunday.
+
+## Obsolete segments
+
+None yet.
+
+## Writing queries with segments
```
jmccrosky: I think we should at least mention somewhere the potential problem that a ping with a given submission_date may be talking about activity that happened on an earlier date.

src/concepts/segments.md
felixlawrence: Added "which would raise our suspicions", which signals the level of certainty that people should take from such a discovery - sound good?

src/concepts/segments.md
felixlawrence: I rephrased - is it better?

felixlawrence: > *Semi-regular users v1* is the other users: the segment is defined as 
clients who browsed >=5 URIs on at least two days in *some but not all* of the last four weeks. 
This seems to be a smaller segment of our users, and has relatively high retention.

src/concepts/segments.md
felixlawrence: OK, if we're going to have this debate then to ground it I think I should spend a day writing up a doc that outlines the plausible options, and the pros and cons of each of them. I'll consider usefulness to decision making, ease of interpretation, ease of implementation, and surprise in the pros and cons. As far as I'm aware, there is no clear right answer here - each option involves some ickiness and/or surprise - so if your instincts differ to mine then we should check this out thoroughly.

src/concepts/segments.md
felixlawrence: Can you give a pathological example where that would mislead someone in a segment-specific way?

src/concepts/segments.md
tdsmith: I guess I wonder why I want to know "how many users ever had each label in the last month" and not "of the users we saw in the last month, how many have this label today"? The latter seems more responsive; zero-sum changes due to transitions seem easier to interpret than an inflation of one group followed by a deflation of another.

If these are states that a user can move through, it doesn't really make sense to have a user in a superposition of states. But maybe a state machine is a bad model and maybe that's not what we're plotting.

Which is to say I'll also read this doc with interest!

felixlawrence: [Here's the doc](https://docs.google.com/document/d/1OpejuD-wtViUX4nwddmaIHF85hU8WknjJBT-zjxxgkE/) - it's probably best to have the discussion in it (inline comments and/or writing "X's take" sections at the end).

src/concepts/segments.md
jmccrosky: Looks like you edited further?  In any case, I'm fine with whatever you think is appropriate.

src/concepts/segments.md
jmccrosky: You really don't want to? :P

src/concepts/segments.md
jmccrosky: What's in your comment looks good.  What's in the branch still seems confusing.

src/concepts/segments.md
jmccrosky: If the user uses the browser in one week, closes the browser before a ping is sent, and then reopens a week later for further use, we'll think that the user was inactive in the first week.

src/concepts/segments.md
felixlawrence: In the "Conversation" view, github presents the code as it was when you made the comment (though adds an "Outdated" badge). What's in my comment is in the branch now :)

src/concepts/segments.md
felixlawrence: There's no "outdated" badge on this one because you commented on line 15 then I edited line 14, but yes I updated the source adding "which would raise our suspicions" to line 14

src/concepts/segments.md
felixlawrence: Yep, this is a general hazard of using `submission_date` - although vastly reduced by [pingsender](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/internals/pingsender.html). Since it's a general hazard, it could be mentioned on every page on DTMO that uses `submission_date` - and there are lots of other hazards that could be mentioned on every page too. This is impractical, so let's leave this to the "Common Analysis Gotchas" page that defines `submission_date`, mentions that using it (rather than client timestamps) is the standard for our data analysis and links to why.

I think this is the kind of detail where the people for whom it's going to be an issue (DSes forensically tracking down why numbers are slightly different than expected) should be expected to know about the properties of `submission_date`, and they are likely to expect that a behavioural segment defined in terms of past usage is defined in terms of `submission_date`, with all the consequences that entails. And if they don't expect it, they can track it down in the ETL.

If this was going to mislead stakeholders as they try to use these segments to do things, then yes we should flag those issues. But I can't think of likely/plausible ways for this to happen...

src/concepts/segments.md
```diff
@@ -0,0 +1,134 @@
...
+(FIXME: This point needs special attention in the review process - it appears to be the best option)
+
+
+### Example queries
+
+DAU for regular users v1:
+```lang=sql
+SELECT
+    submission_date,
+    COUNT(DISTINCT client_id) AS dau_regular_v1
```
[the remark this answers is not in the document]
felixlawrence: Eek - maybe you can't figure the other queries out because they're _very wrong_? I can't figure them out either, and they appear to ignore any user not also active on the current day.

I _think_ i intended to replace `clients_daily` with `clients_last_seen`, so as to join `clients_last_seen` on itself. But regardless this is up in the air pending the MAU discussion...

felixlawrence:
I don't think we're going to come to an agreement around MAU quickly. I would like to start getting value from these segments in non-MAU, non-GUD contexts quickly. I'm going to scrub the "how to query MAU" parts from this document and note that the definition of MAU is TBD in the context of segments.

I'm also going to tweak the definitions slightly in line with [advice from jklukas](https://github.com/mozilla/bigquery-etl/pull/825#issuecomment-603589147) - I updated the [original "proposing a trio of segments" gdoc](https://docs.google.com/document/d/1eI7dXB6iYl_8gQ0wXdPiilLDe4bzAShdgigrDCqThvg/) with the latest plan for the definition.

src/concepts/segments.md
jmccrosky: OK, fair enough :)

felixlawrence:
Thanks! Will do the "beta" comment.

I didn't give approximate numbers because:
1. They'll go out of date
2. I wasn't sure whether the data (specifically retention, but also DAU for the segments themselves) is public or NDA and didn't want to breach that.

Point 1 is easy to solve - give round numbers and add "as of March 2020" - but I'll need to research point 2.

src/concepts/segments.md
```diff
@@ -0,0 +1,105 @@
...
+Observationally, on any given day this segment seems to contain a large fraction of our users 
+and has exceptionally high retention.
+
+*New/irregular users v1* is defined as 
+clients who did not browse >=5 URIs on any of the previous 6 days, and browsed >=5 URIs on _less than two_ days in each of the three weeks before that.
+This is a smaller segment of our users and has low retention 
+(though "activation" is likely a more relevant word than "retention" for many of these clients).
+
+*Semi-regular users v1* contains the other clients.  
+This seems to be a smaller segment of our users, and has relatively high retention.
```
jmccrosky: I thought you were going to add "We do not yet know what proportion of users in this segment stay in this segment for an extended period, and what proportion are in transition between other segments." here?

PR #432, opened 2020-03-30
Typo in nondesktop product query

jklukas:


jklukas:
Going ahead and merging without review since it's a simple typo.

PR #459, opened 2020-04-22
Add documentation for search_clients_last_seen

fbertsch:


src/datasets/search/search_clients_last_seen/intro.md
```diff
@@ -0,0 +1,83 @@
...
+
+`udf.bits_to_days_seen` - The number of days the user was active.
+`udf.bits_to_days_since_seen` - The number of days since the user was last active.
+`udf.bits_to_days_since_first_seen` - The number of days since the user's _first_
+active day. Note that this will be at most 365 days, since that is the beginning
+of history for this dataset.
+`udf.bits_to_active_n_weeks_ago` - Returns whether or not the user was active n weeks
+ago for the given activity type.
+
+#### Engine Searches
```
benmiroglio, line 58: Maybe add here that this dataset was designed for LTV / revenue matching. This field is especially "weird" so adding some context as to why would be useful.

PR #462, opened 2020-04-27
Docs for bit pattern UDF interface

jklukas:
~~Most of the UDFs referenced here don't actually exist yet. The idea is to document bit pattern usage assuming these exist in order to gather feedback and update the design before putting the effort into implementation.~~

These UDFs now exist and are published.

### DIFF
--- a/src/cookbooks/clients_last_seen_bits.md
+++ b/src/cookbooks/clients_last_seen_bits.md
@@ -1,9 +1,19 @@
 # Working with Bit Patterns in Clients Last Seen
 
+Monthly active users (MAU) is a windowed metric that requires joining data
+per client across 28 days. Calculating this from individual pings or daily
+aggregations can be computationally expensive, which motivated creation of the
+[`clients_last_seen` dataset](../datasets/bigquery/clients_last_seen/reference.md)
+for desktop Firefox and similar datasets for other applications.
+
 A powerful feature of the `clients_last_seen` methodology is that it doesn't
 record specific metrics like MAU and WAU directly, but rather each row stores
 a history of the discrete days on which a client was active in the past 28 days.
-This history is encoded as a "bit pattern" where the physical
+We could calculate active users in a 10-day or 25-day window just as efficiently
+as a 7-day (WAU) or 28-day (MAU) window. But we can also define completely new
+metrics based on these usage histories, such as various retention definitions.
+
+The usage history is encoded as a "bit pattern" where the physical
 type of the field is a BigQuery 64-bit integer, but logically the integer
 represents an array of bits, with each 1 indicating a day where the given clients
 was active and each 0 indicating a day where the client was inactive. This
@@ -23,8 +33,8 @@ and wasn't seen in any of the 27 days previous. A value of `2` means
 the client was seen 1 day ago, but not on `submission_date`. A value of `3`
 means that the client was seen on `submission_date` _and_ the day previous.
 
-It's much easier to reason about these bit patterns when we view them
-as strings of ones and zeros, so we've provided a UDF to convert these
+It's much easier to reason about these bit patterns, however, when we view them
+as strings of ones and zeros. We've provided a UDF to convert these
 values to "bit strings":
 
 ```
@@ -43,8 +53,8 @@ right to left in the string of bits, the "lowest" to bits are set (1) while
 the rest of the bits are unset (0).
 
 Let's consider a larger value `8256`. In terms of powers of two, this is equal
-to `2^13 + 2^6`, so we would expect its string representation to have two `1`
-values. If we label the rightmost bit as "offset 0", we would expect the set
+to `2^13 + 2^6` and its string representation should have two `1` values.
+If we label the rightmost bit as "offset 0", we would expect the set
 bits to be at offsets `-13` and `-6`:
 
 ```
@@ -62,12 +72,21 @@ SELECT udf.bits28_from_string('0000000000000010000001000000')
 >>> 8256
 ```
 
-The leading zeros are optional for this function:
+Note that the leading zeros are optional for this function:
 
 ```
 SELECT udf.bits28_from_string('10000001000000')
 
 >>> 8256
+```
+
+Finally, we can translate this into an array of concrete dates by passing
+a value for the date that corresponds to the rightmost bit:
+
+```
+SELECT udf.bits28_to_dates(8256, '2020-01-28')
+
+>>> ['2020-01-15', '2020-01-22']
 ```
 
 ### Why 28 bits instead of 64?
@@ -77,11 +96,11 @@ so we could technically store 64 days of history per bit pattern. Limiting
 to 28 bits is a practical concern related to storage costs and reprocessing concerns.
 
 Consider a client that is only active on a single day and then never shows up again.
-
 A client that becomes inactive will eventually fall outside the 28-day usage
-window and will thus not have a row in following days of `clients_last_seen`.
+window and will thus not have a row in following days of `clients_last_seen`
+and we no longer duplicate that client's data for those days.
 
-Tables following the `clients_last_seen` methodology have to be populated
+Also, tables following the `clients_last_seen` methodology have to be populated
 incrementally. For each new day of data, we have to reference the previous
 day's rows in `clients_last_seen`, take the trailing 27 bits of each pattern
 and appending a 0 or 1 to represent whether the client was active in the new
@@ -126,7 +145,7 @@ to be necessary for consistency in how we define various retention metrics.
 Consider if we wanted to compare 1-week, 2-week, and 3-week retention metrics on a
 single plot. If we use forward-looking windows, then the point labeled 2020-01-01
 describes the same set of users for all three metrics and how their activity
-differs over time. If we used backwards-looking windows, then each of these three
+differs over time. If we use backwards-looking windows, then each of these three
 metrics is considering a separate population of users.
 We'll discuss this in more detail later.
 
@@ -152,8 +171,8 @@ Let's dive more deeply into that bit string representation:
     0  0  0  0  0  0  0  0  0  0  0  0  0  0  1  0  0  0  0  0  0  1  0  0  0  0  0  0
     ──────────────────────────────────────────────────────────────────────────────────
     │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │
-    │ 26  │ 24  │ 22  │ 20  │ 18  │ 16  │ 14  │ 12  │ 10  │  8  │  6  │  4  │  2  │  0
-   27    25    23    21    19    17    15    13    11     9     7     5     3     1
+    │-26  │-24  │-22  │-20  │-18  │-16  │-14  │-12  │-10  │ -8  │ -6  │ -4  │ -2  │  0
+  -27   -25   -23   -21   -19   -17   -15   -13   -11    -9    -7    -5    -3    -1
    
   └──────────────────────────────────────────────────────────────────────────────────┘
   MAU                                                             └──────────────────┘
@@ -172,7 +191,7 @@ We provide a special function to tell us how many days have elapsed since
 the most recent activity encoded in a bit pattern:
 
 ```
-SELECT udf.bits28_days_since_seen(8256)
+SELECT udf.bits28_days_since_seen(udf.bits28_from_string('10000001000000'))
 
 >>> 6
 ```
@@ -201,8 +220,8 @@ pattern for this client will look like:
     0  0  0  0  0  0  0  0  0  0  0  0  0  1  0  0  0  0  0  0  1  0  0  0  0  0  0  0
     ──────────────────────────────────────────────────────────────────────────────────
     │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │  │
-    │ 26  │ 24  │ 22  │ 20  │ 18  │ 16  │ 14  │ 12  │ 10  │  8  │  6  │  4  │  2  │  0
-   27    25    23    21    19    17    15    13    11     9     7     5     3     1
+    │-26  │-24  │-22  │-20  │-18  │-16  │-14  │-12  │-10  │ -8  │ -6  │ -4  │ -2  │  0
+  -27   -25   -23   -21   -19   -17   -15   -13   -11    -9    -7    -5    -3    -1
    
   └──────────────────────────────────────────────────────────────────────────────────┘
   MAU                                                             └──────────────────┘
@@ -210,15 +229,14 @@ pattern for this client will look like:
                                                                                    DAU
 ```
 
-The entire pattern has simply shifted one offset to the right, with the leading zero
+The entire pattern has simply shifted one offset to the left, with the leading zero
 falling off (since it's now outside the 28-day range) and a trailing zero added
 on the right (this would be a `1` instead if the user had been active on 2020-01-29).
 
-The `days_seen_bits` value is now `16512` and the
-`days_since_seen` value is now `7`, which is outside the WAU window:
+The `days_since_seen` value is now `7`, which is outside the WAU window:
 
 ```
-SELECT udf.bits28_days_since_seen(16512)
+SELECT udf.bits28_days_since_seen(udf.bits28_from_string('100000010000000'))
 
 >>> 7
 ```
@@ -231,9 +249,10 @@ portion of clients active on 2020-01-01 are still active some number of days
 later.
 
 When we were talking about backward-looking windows, our anchor date or "day 0"
-was always the `submission_date`. When we define forward-looking windows, however,
-we always choose an anchor date some time in the past. How we number the individual
-bits depends on what anchor date we choose.
+was always the most recent day, corresponding to the rightmost bit.
+When we define forward-looking windows, however, we always choose an anchor date
+some time in the past. How we number the individual bits depends on what
+anchor date we choose.
 
 For example, in [GUD](../tools/gud.md), we show a "1-Week Retention" which considers a window of 14 days.
 For each client active in "week 0" (days 0 through 6), we determine retention by
@@ -256,8 +275,9 @@ the array to define our new "day 0" which corresponds to 2020-01-15:
 ```
 
 This client has a bit set in both week 0 and in week 1, so logically this client
-can be considered retained. They should be counted in both the denominator and
-in the numerator for the "1-Week Retention" value on 2020-01-15.
+can be considered retained; they should be counted in both the denominator and
+in the numerator for the "1-Week Retention" value on 2020-01-15. But how can
+we extract this usage per week information programmatically?
 
 Extracting the bits for a specific week can be achieved via UDF:
 
@@ -314,6 +334,9 @@ The `bits28_retention` struct also has `day_21` and `day_28` fields that can
 be used to calculate "2-Week Retention" and "3-Week Retention".
 
 ### Full example query for 1-Week Retention variants
+
+Finally, we can put these UDFs to use by aggregating numerators and denominators
+to calculate real retention metrics:
 
 ```
 WITH base AS (
--- a/src/cookbooks/clients_last_seen_bits.md
+++ b/src/cookbooks/clients_last_seen_bits.md
@@ -407,22 +407,22 @@ SET n = 3;
 WITH base AS (
   SELECT
     *,
-    udf.bits28_range(days_seen_bits, -n + 0, 1) AS seen_on_day_0,
-    udf.bits28_range(days_seen_bits, -n + 1, n - 1) AS seen_after_day_0
+    udf.bits28_active_in_range(days_seen_bits, -n + 0, 1) AS seen_on_day_0,
+    udf.bits28_active_in_range(days_seen_bits, -n + 1, n - 1) AS seen_after_day_0
   FROM
     telemetry.clients_last_seen )
 SELECT
-  DATE_SUB(submission_date, INTERVAL n DAYS) AS metric_date,
+  DATE_SUB(submission_date, INTERVAL n DAY) AS metric_date,
   SAFE_DIVIDE(
     COUNTIF(seen_on_day_0 AND seen_after_day_0),
     COUNTIF(seen_on_day_0)
   ) AS retention_n_day
 FROM
   base
-GROUP BY
-  metric_date
 WHERE
   submission_date = '2020-01-28'
+GROUP BY
+  metric_date
 ```
 
 ### Retention using activity date
@@ -455,22 +455,22 @@ SET offset_to_day_0 = -n - cushion_days;
 WITH base AS (
   SELECT
     *,
-    udf.bits28_range(days_seen_session_start_bits, offset_to_day_0 + 0, 1) AS seen_on_day_0,
-    udf.bits28_range(days_seen_session_start_bits, offset_to_day_0 + 1, n - 1) AS seen_after_day_0
+    udf.bits28_active_in_range(days_seen_session_start_bits, offset_to_day_0 + 0, 1) AS seen_on_day_0,
+    udf.bits28_active_in_range(days_seen_session_start_bits, offset_to_day_0 + 1, n - 1) AS seen_after_day_0
   FROM
     org_mozilla_fenix.baseline_clients_last_seen )
 SELECT
-  DATE_SUB(submission_date, INTERVAL offset_to_day_0 DAYS) AS metric_date,
+  DATE_SUB(submission_date, INTERVAL offset_to_day_0 DAY) AS metric_date,
   SAFE_DIVIDE(
     COUNTIF(seen_on_day_0 AND seen_after_day_0),
     COUNTIF(seen_on_day_0)
   ) AS retention_n_day
 FROM
   base
-GROUP BY
-  metric_date
 WHERE
   submission_date = '2020-01-28'
+GROUP BY
+  metric_date
 ```
 
 ## UDF Reference
--- a/src/cookbooks/clients_last_seen_bits.md
+++ b/src/cookbooks/clients_last_seen_bits.md
@@ -522,6 +522,8 @@ active_in_week_0_after_metric_date = false
 
 ### N-day Retention
 
+_Not an official metric. This section is intended solely as an example of advanced usage._
+
 As an example of a novel metric that can be defined using the low-level
 bit pattern UDFs, let's define `n`-day retention as the fraction of clients active on a given day
 who are also active within the next `n` days. For example, 3-day retention would
@@ -544,6 +546,8 @@ WITH base AS (
     telemetry.clients_last_seen )
 SELECT
   DATE_SUB(submission_date, INTERVAL n DAY) AS metric_date,
+
+  -- NOT AN OFFICIAL METRIC
   SAFE_DIVIDE(
     COUNTIF(seen_on_day_0 AND seen_after_day_0),
     COUNTIF(seen_on_day_0)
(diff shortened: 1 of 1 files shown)


src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,136 @@
+# Working with Bit Patterns in Clients Last Seen
+
+An powerful feature of the `clients_last_seen` methodology is that we can capture
```
wlach, line 3:
```suggestion
A powerful feature of the `clients_last_seen` methodology is that we can capture
```

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,136 @@
+# Working with Bit Patterns in Clients Last Seen
```
wlach, line 1: You may already have been thinking about this, but I think we need some kind of introductory paragraph(s) to motivate this content. Describing some concrete questions that might be answered by this material would be good.

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,136 @@
...
+In this picture, we've annotated the windows for DAU (day 0 only), 
+WAU (days 0 through -6) and MAU (days 0 through -27). This particular client
+won't count toward DAU for 2020-01-28, but the client does count towards both
+WAU and MAU.
+
+You may notice that the day 13 bit doesn't end up affecting these metrics.
+For DAU, WAU, and MAU, all that matters is the rightmost bit that's set,
+telling us about the most recent activity seen for that user. We provide
+a special function for that use case:
+
```
wlach, line 61:
```suggestion
```

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,136 @@
...
+## Retention: Forward-looking windows
+
+For retention calculations, we user forward-looking windows. This means that
+when we report a retention value for 2020-01-01, we're talking about what
+portion of clients active on 2020-01-01 are still active some number of days
+later.
+
+When we were talking about backward-looking windows, our anchor date or "day 0"
+was always the `submission_date`. When we define forward-looking windows, however,
```
wlach, line 114: What does the `submission_date` actually means here? I assume you mean `submission_date` of clients_last_seen?

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,136 @@
...
+we always choose an anchor date some time in the past. How we number the individual
+bits depends on what anchor date we choose.
+
+In GUD, we show a "Week 1 Retention" which considers a window of 14 days.
```
wlach, line 118:
```suggestion
In [GUD](../tools/gud.md), we show a "Week 1 Retention" which considers a window of 14 days.
```

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,136 @@
...
+
+Let's consider the same user from before, grabbing the `days_seen_bits` value
+with `submission_date = 2020-01-28`. We'll still choose a window that _ends_
+on the submission date, but the anchor date (day 0) will now be 13 days earlier:
+
```
wlach, line 123:
```suggestion
```

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,136 @@
...
+    0  0  0  0  0  0  0  0  0  0  0  0  0  0  1  0  0  0  0  0  0  1  0  0  0  0  0  0
+    ──────────────────────────────────────────────────────────────────────────────────
+                                              │  │  │  │  │  │  │  │  │  │  │  │  │  │
+                                              │  1  │  3  │  5  │  7  │  9  │ 11  │ 13
+                                              0     2     4     6     8    10    12                                     
+                                             └───────────────────┘
+                                                    Week 0       └───────────────────┘
+                                                                        Week 1
+```
+
```
wlach, line 135: I think some kind of explanation of what we just determined by this calculation would be helpful here, instead of relying on the reader to fill in the blanks.

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,136 @@
...
+backward-looking windows. This is what we do for our canonical _usage_ measures
+DAU, WAU, and MAU.
+
+Let's imagine a single row from `clients_last_seen` with `submission_date = 2020-01-28`.
+The `days_seen_bits` field records usage for a single client over a period of 28 days
+_ending_ on the given `submission_date`. We will call this anchor date "day 0" 
+(2020-01-28 in this case) and count backwards to "day -27" (2020-01-01).
+
+Let's suppose this client was only active on two days in the past month:
+2020-01-22 (day -6) and 2020-01-15 (day -13). That client's`days_seen_bits`
```
wlach, line 28:
```suggestion
2020-01-22 (day -6) and 2020-01-15 (day -13). That client's `days_seen_bits`
```

src/cookbooks/clients_last_seen_bits.md
jklukas: Yes, I'll try to make this more clear.

src/cookbooks/clients_last_seen_bits.md
jklukas: Will do.

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,306 @@
...
+  ) AS retention_n_day
+FROM
+  base
+GROUP BY
+  metric_date
+WHERE
+  submission_date = '2020-01-28'
+```
+
+### Retention using activity date
```
jklukas, line 258: This is getting fairly speculative here. I've never actually run any queries using these activity date-based bit patterns, so they're unvetted. But I'm fairly confident that the kind of calculation described here is feasible.

src/cookbooks/clients_last_seen_bits.md
jmccrosky: One could argue that we should not use the name `submission_date` at all in these tables, since this is far enough from raw pings that there is no reasonable definition of when anything was "submitted".  I haven't thought deeply about this though, so feel free to disagree :)

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,375 @@
...
+later.
+
+When we were talking about backward-looking windows, our anchor date or "day 0"
+was always the `submission_date`. When we define forward-looking windows, however,
+we always choose an anchor date some time in the past. How we number the individual
+bits depends on what anchor date we choose.
+
+For example, in [GUD](../tools/gud.md), we show a "1-Week Retention" which considers a window of 14 days.
+For each client active in "week 0" (days 0 through 6), we determine retention by
+checking if they were also active in "week 1" (days 0 through 13).
```
jmccrosky, line 119: shouldn't this be "(days 7 through 13)"?

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,136 @@
+# Working with Bit Patterns in Clients Last Seen
+
+An powerful feature of the `clients_last_seen` methodology is that we can capture
+per-day usage information for the entire 28-day period that each row in the table
```
magorlick, line 4: This reads a little bit confusing.  How can it be per day and for an entire 28 day period in one row? We are doing a rolling 28 day aggregation here that is tied to a day?  I thought it was fairly flexible to roll up info that occurred over the last 28 days in whatever way you want (could be less than 28 day roll up you just don't have access to things occurring outside of that window?  I always like to call out exactly what makes each row unique when I describe a table.

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,136 @@
...
+describes. We encode usage information in "bit patterns" where the physical
+type of the field is a BigQuery 64-bit integer, but logically the integer
+represents an array of bits, with each 1 indicating a day where the given clients
```
magorlick, line 7: How is the array ordered?

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,136 @@
...
+was active and each 0 indicating a day where the client was inactive.
+
+The representation is space-efficient and actually easier to manipulate
+in BigQuery than a date array would be, but requires some practice to interact
+with directly. For well-established usage and retention metrics, we generally
+provide logic in views to insulate users from having to ever consider the
+raw bit patterns, but new investigations may require novel retention definitions.
+
+## Usage: Backward-looking windows
```
magorlick, line 16: Don't we look forward in GUD?  For the record I prefer backward.  It raises less questions from folks that wonder where the recent data is.

src/cookbooks/clients_last_seen_bits.md
magorlick: Also calling this out as a limitation of clients last seen (only 28 days) is useful.

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,136 @@
...
+
+The simplest application of usage bit patterns is for calculating metrics in
+backward-looking windows. This is what we do for our canonical _usage_ measures
+DAU, WAU, and MAU.
+
+Let's imagine a single row from `clients_last_seen` with `submission_date = 2020-01-28`.
+The `days_seen_bits` field records usage for a single client over a period of 28 days
```
magorlick, line 23: Any usage criteria baked in here? Does that limit this approach?

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,136 @@
...
+
+```
+SELECT udf.bits28_to_bitstring(32768)
+-- 0000000000000010000001000000
+```
+
+Let's make that representation a bit more visual:
+
+```
+    0  0  0  0  0  0  0  0  0  0  0  0  0  0  1  0  0  0  0  0  0  1  0  0  0  0  0  0
```
magorlick, line 40: nice

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,136 @@
...
+                                                                   WAU             └─┘
+                                                                                   DAU
+```
+
+In this picture, we've annotated the windows for DAU (day 0 only), 
+WAU (days 0 through -6) and MAU (days 0 through -27). This particular client
+won't count toward DAU for 2020-01-28, but the client does count towards both
+WAU and MAU.
+
+You may notice that the day 13 bit doesn't end up affecting these metrics.
```
magorlick, line 57: It does affect MAU, but the activity on day 6 also affects MAU so you don't 'need' to see them on the 13th to count them in this specific case.

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,136 @@
...
+For DAU, WAU, and MAU, all that matters is the rightmost bit that's set,
```
magorlick, line 58: You totally lost me here.  Imagine I have never studied bits before.  What do you mean by the rightmost bit? 0 on day 0 or 6 from the inscrutable 8256?

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,136 @@
...
+  └──────────────────────────────────────────────────────────────────────────────────┘
+  MAU                                                             └──────────────────┘
+                                                                   WAU             └─┘
+                                                                                   DAU
+```
+
+The `days_seen_bits` value is now `16512` and the user is indeed outside the WAU window:
+
+```
+SELECT udf.bits28_days_since_seen(16512)
```
magorlick, line 101: This is still a mystery to me but maybe it is ok if folks don't understand how you go from the inscrutable number to 7 days since last seen and can just rely on the UDF

src/cookbooks/clients_last_seen_bits.md
magorlick: I prefer more specificity.  If these were all derived from submission dates that makes sense to me. If they are all different kinds of dates that were processed downstream to fit into a model of the truth we can use some other word.  I have a strong aversion to just using 'date' because it is a SQL function and I have to start putting quotes on it all the time.

src/cookbooks/clients_last_seen_bits.md
jklukas: WAU and MAU are backward-looking metrics where the value for 2020-01-28 depends on activity from 2020-01-01 to 2020-01-28.

Retention metrics are forward-looking where the value for 2020-01-28 depends on activity happening _after_ that date.

This is confusing and subtle for those who have to calculate the metrics, but I do think it's the most straight-forward approach from the consumer side. For a given date, you may want to compare 3-day retention, 1-week retention, and 2-week retention; this only makes sense if we align them based on the first day of the window.

src/cookbooks/clients_last_seen_bits.md
jklukas: I will need to add some discussion of usage criteria, but yes usage criteria have to be fixed as part of the `clients_last_seen` where each usage criterion is its own bit pattern field. It's a known deficiency right now that we don't have a defined workflow for how DS can quickly test out new usage criterion definitions and propose them for inclusion. We need to do a full backfill to add a new usage criterion.

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,375 @@
...
+```
+
+The `bits28_range` function is powerful and flexible, but a bit cumbersome.
+We provide a `bits28_retention` convenience function that returns a nested
+structure with the standard retention definitions precalculated:
+
+```
+SELECT
+  submission_date,
+  udf.bits28_retention(days_seen_bits, submission_date).day_13.*
```
jklukas, line 170: I'd appreciate thoughts on the interface for this `bits28_retention` function. I wanted to make it easy to use and in particular wanted to limit the number of parameters that you have to pass in. The idea of having this return a whole struct of the various retention metrics is a bit weird, but hopefully is convenient?

Is it clear from the description how to use this?

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,375 @@
...
+>> 1
+```
+
+
+### bits28_range
+
+Return an INT64 representing a range of bits from a source bit pattern.
+
+```
+bits28_range(bits INT64, offset_to_day_0 INT64, start_bit INT64, number_of_bits INT64)
```
jklukas, line 358: This ` bits28_range` function is the trickiest and has the highest margin for error. It has a gaggle of params, all of which are ints, so ripe for getting the order wrong, etc.

This function also kind of does a lot. We could instead express this as two separate functions, one that trims off bits from the front and one that trims off bits from the end, something like:

```
-- Get bits 7 though 13.
bits28_take_first(bits28_take_last(bits, 14), 7)
```

jklukas: I'm thinking that potentially a more simplistic function would actually be easier to reason about if we provide sufficient documentation on usage.

We could have `bits28_range(bits, start_index, number_of_bits)` where the index is a negative number that indicates a number of bits from the end of the pattern. So dealing with a 2-week window could look like:

```
SELECT
  DATE_SUB(submission_date, INTERVAL 13 DAY) AS metric_date,
  BIT_COUNT(udf.bits28_range(days_seen_bits, -13 + 0, 1)) > 0 AS active_on_metric_date,
  BIT_COUNT(udf.bits28_range(days_seen_bits, -13 + 0, 7)) > 0 AS active_in_week_0,
  BIT_COUNT(udf.bits28_range(days_seen_bits, -13 + 1, 6)) > 0 AS active_in_week_0_after_metric_date,
  BIT_COUNT(udf.bits28_range(days_seen_bits, -13 + 7, 7)) > 0 AS active_in_week_1,
```

src/cookbooks/clients_last_seen_bits.md
jklukas: I'm hoping this is addressed by the new "Understanding bit patterns" section, but let me know if it still feels like that needs to be addressed here.

src/cookbooks/clients_last_seen_bits.md
jklukas: I added a new "Forward-looking windows and backward-looking windows" section above this to hopefully capture the gotchas here.

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,487 @@
...
+was active and each 0 indicating a day where the client was inactive. This
+article discusses the details of how we represent usage in bit patterns,
+how to extract standard usage and retention metrics,
+and how to build new metrics from them.
+
+## Understanding bit patterns
+
+If you look at the `days_seen_bits` field in `telemetry.clients_last_seen`,
+you'll see seemingly random whole numbers, some as large as nine digits.
+How can we interpret these.
```
hamilton, line 18:
```suggestion
How should we interpret these?
```

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,487 @@
...
+
+This client has a bit set in both week 0 and in week 1, so logically this client
+can be considered retained. They should be counted in both the denominator and
+in the numerator for the "1-Week Retention" value on 2020-01-15.
+
+Extracting the bits for a specific week can be achieved via UDF:
+
+```
+SELECT
+  -- Signature is bits28_range(offset_to_day_0, start_bit, number_of_bits)
```
fbertsch, line 257: This says 3 args, but below has 4 args.

jklukas, line 257: This is an error right now.

jklukas:
These UDFs are all now implemented in https://github.com/mozilla/bigquery-etl/pull/948 and I've gone ahead and published them so that folks can start playing with them. I've also fixed various issues in the text here to make sure the examples match the published UDFs.

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,539 @@
...
+
+```
+SELECT
+  submission_date,
+  udf.bits28_retention(days_seen_bits, submission_date).day_13.*
+FROM
+  telemetry.clients_last_seen
+
+/*
+| submission_date | metric_date | active_on_day_0 | active_in_week_0 | active_in_week_1 |
```
fbertsch, line 328: This seems to be missing `active_in_week_0_after_metric_date`. What is that used for?

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,539 @@
...
+SELECT
+  udf.bits28_active_in_range(days_seen_bits, -13 + 0, 7) AS active_in_week_0
+  udf.bits28_active_in_range(days_seen_bits, -13 + 7, 7) AS active_in_week_1
+FROM
+  telemetry.clients_last_seen
+```
+
+The `bits28_range` function is powerful and flexible, but a bit cumbersome.
+We provide a `bits28_retention` convenience function that returns a nested
+structure with the standard retention definitions already calculated:
```
fbertsch, line 318: Is `standard retention definitions already calculated` the correct term here? I believe it should be `standard activity definitions already calculated`. We use these activity definitions to derive retention.

src/cookbooks/clients_last_seen_bits.md
jklukas: I added an additional field to the full retention query example that uses this field.

jklukas:
@felixlawrence - I'd be interested in your review here. Is this documentation sufficient that you'd feel confident being able to express your recent segmentation definitions using these UDFs? Does this interface feel more usable than existing UDFs?

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,553 @@
...
+>> 1
+```
+
+
+### `bits28_range`
+
+Return an INT64 representing a range of bits from a source bit pattern.
+
+```
+bits28_range(bits INT64, offset INT64, n_bits INT64)
```
felixlawrence, line 528: These methods seem to aim to remove the need for users to do bitwise operations explicitly. This is probably sensible: the learning curve is steep (though also short). Though these methods are harder to debug: I can't just print out the string representation of the mask to check that I'm doing what I expected.

It took me a lot of thinking to work out whether `offset` is with respect to the leftmost or rightmost bit, and the direction in which `n_bits` counts.

Given the rest of the docs, I would expect this to be backwards-looking: I would expect `n_bits` to count leftwards and an `offset` of `0` to represent the inclusion of the rightmost bit.

But from the examples it _seems_ that `n_bits` counts rightwards? So this function is most useful for the forwards-looking usecase? Do we expect the forwards-looking usecase to be more common? Should we have separate UDFs for forwards and backwards looking usecases?

As someone most familiar with Python, I (selfishly) want to think about this stuff in terms of slicing a flipped list, with arguments `start` and `stop`. So these examples would read `udf.bits28_range(18, 0, 6)`, `udf.bits28_range(18, 4, 6)` and `udf.bits28_range(18, 0, 4)` respectively.

jklukas: > But from the examples it seems that n_bits counts rightwards? So this function is most useful for the forwards-looking usecase? Do we expect the forwards-looking usecase to be more common? Should we have separate UDFs for forwards and backwards looking usecases?

You are correct on all accounts. For backwards-looking use cases, `bits28_days_since_seen` is the only UDF needed. `bits28_range`, `bits28_active_in_range`, and `bits28_retention` are all intended for forward-looking windows, so they expect a negative offset to determine how many bits left to move before it starts reading, then `n_bits` measures how many bits to the right to move.

jklukas: > As someone most familiar with Python, I (selfishly) want to think about this stuff in terms of slicing a flipped list, with arguments start and stop

What you describe is definitely more straight-forward in terms of how the parameters relate to the bit pattern. But that interface doesn't seem helpful for thinking about forward-looking windows where I think the right thing to do is to define a starting bit and then count rightward.

Does that sound reasonable? I can certainly add more to the documentation of this function to clarify some of the points you found surprising.

felixlawrence: > For backwards-looking use cases, bits28_days_since_seen is the only UDF needed.

Not for my backwards-looking use cases! :sweat_smile:

Since the text above already discusses both forward and backward windows at length, it probably makes sense to specify which of these UDFs are counting forward and which are counting backward. At least here in the docs; optionally also in the naming of the functions, e.g. `bits28_forward_range()`.

felixlawrence: > Does that sound reasonable?

Yes! As long as it's clear that they're for forward looking windows

jklukas: Well, we could modify this function to allow negative numbers for `n_bits`. Negative `n_bits` would mean that we count left from the start offset (backward-looking windows) while positive `n_bits` would mean counting right (forward-looking windows).

felixlawrence: For range specification/slicing, `(start, stop)` is an established format, as is `(start, length)` for positive lengths. I don't think I've seen `(start, length)` overloaded to become `(stop, negative_length)` when `length < 0`.

I would be a bit too scared to use it and will probably stick with masks (I'm OK with that - past the learning curve now) - I would worry about messing up the arithmetic and doing off-by-one miscalculations that take me to the 29th bit, yielding plausible but wrong results.

jklukas: > I would worry about messing up the arithmetic and doing off-by-one miscalculations that take me to the 29th bit

We already have validation logic in this UDF such that we'll raise an error `'Reading %i bits from starting_offset %i exceeds end of bit pattern in call bits28_range(%i, %i, %i)'`. We would add additional checks if we enabled negative `n_bits` to at least catch the "29th bit" case.

> For range specification/slicing, (start, stop) is an established format, as is (start, length) for positive lengths. I don't think I've seen (start, length) overloaded to become (stop, negative_length) when length < 0.

I feel you here. I'm not sure I've ever seen an API that's quite analogous to what we have here. To be logically consistent for forward and backward-looking cases, I think we need to be consistent in thinking of offsets in terms of higher values as we move to the right. But maybe you're right that it would be easier to reason about if we had separate `bits28_range_forward` and `bits28_range_backward` or `bits28_range_right` and `bits28_range_left`.

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,563 @@
...
+
+```
+DECLARE n INT64;
+SET n = 3;
+
+WITH base AS (
+  SELECT
+    *,
+    udf.bits28_active_in_range(days_seen_bits, -n + 0, 1) AS seen_on_day_0,
+    udf.bits28_active_in_range(days_seen_bits, -n + 1, n - 1) AS seen_after_day_0
```
jklukas, line 411:
```suggestion
    udf.bits28_active_in_range(days_seen_bits, -n + 1, 1) AS seen_on_day_0,
    udf.bits28_active_in_range(days_seen_bits, -n + 2, n - 1) AS seen_after_day_0
```

I think I made a mistake here.

src/cookbooks/clients_last_seen_bits.md
magorlick: Yes, this is clear below.

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,689 @@
...
+ORDER BY
+  submission_date
+```
+
+Under the hood, `days_since_seen` is calculated using the `bits28_days_since_seen`
+UDF which is explained in more detail later in this article.
+
+Note that the desktop `clients_last_seen` table also has additional bit pattern
+fields corresponding to additional [usage criteria](../metrics/index.md),
+so other variants on MAU can be calculated like:
```
magorlick, line 61: Do you want to put forward any comments on what it takes to get new usage critieria? Why these usage criteria?  My understanding is that this criteria is not flexible in production as is, instead we need to make a formal request for each additional usage criteria.

src/cookbooks/clients_last_seen_bits.md
magorlick: I like the discussion above.  I would call out the challenges to add new usage criteria as noted.

src/cookbooks/clients_last_seen_bits.md
magorlick: Do we want to keep this or do we want to fairly firmly be in the submission_date camp?

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -0,0 +1,689 @@
...
+```
+DECLARE n, cushion_days, offset_to_day_0 INT64;
+SET n = 3;
+SET cushion_days = 2;
+SET offset_to_day_0 = -n - cushion_days;
+
+WITH base AS (
+  SELECT
+    *,
+    udf.bits28_active_in_range(days_seen_session_start_bits, offset_to_day_0 + 0, 1) AS seen_on_day_0,
```
magorlick, line 584: Do we need to update this query as well?

    udf.bits28_active_in_range(days_seen_session_start_bits, offset_to_day_0+ 1, 1) AS seen_on_day_0,
    udf.bits28_active_in_range(days_seen_session_start_bits, offset_to_day_0 + 2, n - 1) AS seen_after_day_0

src/cookbooks/clients_last_seen_bits.md
jklukas: I want to have some advanced examples. I've added more of a disclaimer here about how it's unofficial.

jklukas:
Looking for a final read-through at this point and a concrete approving review, then we'll merge this.

PR #470, opened 2020-05-06
Add MPL license

wlach:


wlach:
@elkevorh here's what we were talking about in the meeting :)

PR #471, opened 2020-05-11
Remove links to the Fx-Data-Planning repo.

mreid-moz:
The content from that repository has been migrated to Mana at
https://mana.mozilla.org/wiki/display/DENG

and may soon move to be linked from:
https://mana.mozilla.org/wiki/display/DATA

There is a corresponding issue at https://github.com/mozilla/Fx-Data-Planning/issues/56

### DIFF
--- a/src/tools/projects.md
+++ b/src/tools/projects.md
@@ -154,6 +154,3 @@ starting a new project using anything in this section.
 
 | Name and repo                  | Description                             |
 |--------------------------------|-----------------------------------------|
-| [`Fx-Data-Planning`][planning] | Quarterly goals and internal documentation
-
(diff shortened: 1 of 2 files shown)


src/tools/projects.md
```diff
@@ -154,6 +154,3 @@ starting a new project using anything in this section.
 
 | Name and repo                  | Description                             |
 |--------------------------------|-----------------------------------------|
-| [`Fx-Data-Planning`][planning] | Quarterly goals and internal documentation
```
wlach, line 157: This is now an empty table, which I guess isn't great, but oh well. We should just move this whole page somewhere more appropriate.

PR #474, opened 2020-05-13
Initial skeleton of MSG section.

jmccrosky:
This is just a rough sketch of the outline I'm thinking of.  Wanted to get feedback on the basic structure before sinking too much time into it.


src/metrics/definitions.md
```diff
@@ -0,0 +1,9 @@
+# Metric Definitions
+
+It is important to understand that metrics are often defined in terms of a usage criteria.
```
jklukas, line 3:
```suggestion
It is important to understand that metrics are often defined in terms of a usage criterion.
```

src/metrics/definitions.md
```diff
@@ -0,0 +1,9 @@
...
+
+* A `usage criterion` defines what a metric is measuring.  It generally specifies a product and what specific use of that product qualifies as active.  For example, “Any Fenix Activity” measures activity for all Fenix profiles that send a telemetry pings.  Similarly, “Opened DevTools” means that we measure activity for all Firefox profiles that send a telemetry ping indicating that DevTools was opened.
```
jklukas, line 5:
```suggestion
* A `usage criterion` defines what a metric is measuring.  It generally specifies a product and what specific use of that product qualifies as active.  For example, “Any Fenix Activity” measures activity for all Fenix profiles that have sent a `baseline` ping.  Similarly, “Opened DevTools” means that we measure activity for all desktop Firefox profiles that have sent a telemetry ping indicating that DevTools was opened.
```

src/SUMMARY.md
```diff
@@ -25,6 +25,13 @@
     * [Accessing Glean data](concepts/glean/accessing_glean_data.md)
     * [Glean Debug ping viewer](concepts/glean/debug_ping_view.md)
   * [Growth & Usage Dashboard (GUD)](tools/gud.md)
+* [Metrics](metrics/index.md)
+  * [Overview](metrics/index.md)
```
jklukas, line 29: Do we need this Overview bullet; I assume the parent `[Metrics]` entry is where we'd give the overview. And we'd want to remove the `[Standard Metrics]` link in getting started or replace that with a separate doc that links to the Metrics section.

jmccrosky: Thanks.  I agree that the overview bullet is unnecessary.  And I did intent for the link in getting started to be a brief introduction with a link to this section.  Will make those changes now.

src/SUMMARY.md
```diff
@@ -25,6 +25,12 @@
     * [Accessing Glean data](concepts/glean/accessing_glean_data.md)
     * [Glean Debug ping viewer](concepts/glean/debug_ping_view.md)
   * [Growth & Usage Dashboard (GUD)](tools/gud.md)
+* [Metrics](metrics/index.md)
```
harterrt, line 28: My gut read is to have the following format:

* Metrics
  * Metric 1
  * Metric 2
  * ...
  * Common Slices (Usage Critera and Slicing Dimensions)

I'm not sure if Metrics Standardization and Policy belongs in DTMO. What does that section cover? If it's "how to engage or change these metrics", then let's add it to `metrics/index.md`

harterrt:
Looks great overall. Some nits inline. Thanks for putting this together, Jesse!

src/metrics/definitions.md
```diff
@@ -0,0 +1,9 @@
+# Metric Definitions
```
wlach, line 1: I wonder if this would better go in the terminology document on docs.tmo? We can always split that up into different types of terminology, adding a metrics section.

https://docs.telemetry.mozilla.org/concepts/terminology.html

Maybe better yet, we could add a section per term, so we could do something like like this:

https://docs.telemetry.mozilla.org/concepts/terminology.html#usage-criterion
https://docs.telemetry.mozilla.org/concepts/terminology.html#dimension

Then, when talking about / introducing these terms in the metrics section, we can link back to these terms.

felixlawrence, line 1: This all smells very GUD-centric: it's describing GUD's method for specifying metrics. Does a DTMO page titled "Metric Definitions" need to be more general than this, or are we OK with neglecting metrics that don't fit this format?

src/SUMMARY.md
jmccrosky: Sounds fair.  I might push for separating usage criteria from slicing dimensions though.  I feel that the definitions around when we consider a user to have used our products may be important - especially if we want to move beyond "had the app/browser open", which I think we should do.

For the how to engage or change these metrics, that seems like a much more unusual and specialized task, so wanted a seperate section for it.  Why do you feel it should be in the index.md?

src/metrics/definitions.md
jmccrosky: yup, good point!  will do :)

jmccrosky: Good point.  Part of that is due to some copy and paste history.  I'll revise.  Generally, the GUD metrics are an important class and we definitely want to support first-class concepts of usage criteria, but we should make clear that not all metrics require usage criteria.  Sound fair?

src/SUMMARY.md
harterrt: > For the how to engage or change these metrics, that seems like a much more unusual and specialized task, so wanted a seperate section for it. Why do you feel it should be in the index.md?

Good point. I think the specialization suggests it may not be a good fit for DTMO, maybe we should maintain the docs for changing metrics elsewhere and link from `index.md`

>  I might push for separating usage criteria from slicing dimensions

Could be. Usage criteria feels like an instance of a slicing dimension to me. E.g. I don't see much difference between slicing to "users who saw at least 5 URIs" and "users who used the browser in DE" but I'm open to being convinced otherwise. 

For now, I'd prefer to add as few pages as possible to keep cognitive load low.

src/metrics/definitions.md
felixlawrence: Is this breakdown into "usage criterion", "dimension" and "metric" useful or confusing outside of a GUD context, even if parts are marked as optional?

This breakdown works well for metrics that boil down to "count the users who meet this usage criterion over a time period defined w.r.t. this day, optionally splitting users by some dimension(s)". This is one of my favourite forms of metric: "set some criteria and count users" is very outlier-resistant, and it encourages people to think and in terms of individual users and their experiences, rather than encouraging people to design a car with room for 2.1 kids when the vast majority of families have an integer number of children.

But this breakdown has a narrow scope that excludes lots of metrics, for example, "Average URIs browsed per client per day", "Total searches made per day", or any funnel conversion rate metrics. I know that GUD supports parameterised retention metrics that compute fractions, but even _that_ doesn't fit this breakdown as described.

What value are we trying to get from this document? How can we get that value in a way that doesn't appear to limit us to a subset of the metrics we need to care about?

src/metrics/definitions.md
```diff
@@ -0,0 +1,9 @@
...
+
+It is important to understand that metrics are often defined in terms of a usage criterion.
+
+* A `usage criterion` defines what a metric is measuring.  It generally specifies a product and what specific use of that product qualifies as active.  For example, “Any Fenix Activity” measures activity for all Fenix profiles that have sent a `baseline` ping.  Similarly, “Opened DevTools” means that we measure activity for all desktop Firefox profiles that have sent a telemetry ping indicating that DevTools was opened.
+
+* A `dimension` allows slicing to a subset of profiles according to characteristics of those profiles.  Available dimensions include: country, channel, OS, etc.  A slice is a set of particular values within dimensions, for example, “country is US” is one slice and “country is either US or DE and channel is release” is another.
```
felixlawrence, line 7: Probably need to emphasise the word "slice" in the same way that we emphasise "dimension", "metric" and "usage criterion".

src/metrics/definitions.md
```diff
@@ -0,0 +1,9 @@
...
+
+* A `metric` summarizes activity for a particular day.  It is defined in terms of a `usage criterion` and `slice`.  For example, the `DAU` metric will tell you how many profiles met the `usage criterion` within the `slice` or, more concretely, might tell you how many profiles had any Firefox Desktop activity within Canada.
```
felixlawrence, line 9: > A `metric` summarizes activity for a particular day.

The first few times I read this, I interpreted it as summarising activity _on_ a particular day, which is not the case. A metric can summarise activity over some time period (e.g. 28 days for MAU), and we define that time period with respect to a particular day. Is there some way to make this clearer?

src/SUMMARY.md
felixlawrence: > I don't see much difference between slicing to "users who saw at least 5 URIs" and "users who used the browser in DE" but I'm open to being convinced otherwise.

In my mental model (which is different to GUD's), a metric is "the thing you want to measure" and sometimes you'll want to _add_ a slicing dimension to measure that thing over a subset of the user population.

If we're investigating a change in a metric, we'll want to slice that metric over many dimensions (check that metric just for German users, check it just for users in the regular users v3 segment, etc) to try to find a slice that isolates the change in the metric.

In the GUD model, which is basically what is presented here, a "usage criterion" is part of the metric's definition - swapping out a usage criterion changes the metric itself, rather than splitting the population into two.

Note that this "try lots of slices to find one that isolates the change" requires that the slices to be defined carefully to avoid tautology ("why yes sir, 100% of the increase in people sending pings today was from people who sent pings today"). If they're defined based on behaviour rather than enduring properties (e.g. country) then they should be defined on _past behaviour_, like I've tried to do with the segments work.

jmccrosky: Regarding having the policy docs elsewhere, I'm not really opposed, but feel that it's nice to have an identified home for MSG and this seems like it will be it.  It seems natural that people might expect to find documents related to our policies here - a nice parallel is the docs on contributing to DTMO that are linked at the bottom.  But I don't feel that strongly.  Any particular reason not to have the policy docs here?

Regarding usage criteria, the reason they're important is to be explicit that we are counting something in particular.  When we use MAU right now, we are implicitly saying that we are counting all desktop users that submitted a ping.  Implicit definitions are not a good thing.  The usage critera forces us to be explicit about what we are counting.

And even if we weren't using the GUD data model, we really need a list of usage criteria as a first-class MSG citizen.  One of the problems we explicitly want to tackle is understand what we should count as an "active user".   This flared up recently with glean.  Even if we allow "any ping" to be the standard usage criteria for a product, it should be explicit.

I absolutely agree that this space should properly support non-GUD metrics.  I'll try to make clearer that the usage criteria section is not necessarily a component of every metric.

src/metrics/definitions.md
jmccrosky: Looking at this now, I think it might be useful to explain what a metric, usage criteria, and slicing dimension is "in situ".  It may not be obvious to people how to use this section.  Perhaps we can have short notes here and also more thorough definitions in the terminology doc?

src/SUMMARY.md
jmccrosky: Will note for @felixlawrence that I have dramatically changed definitions.md to make clear that usage criteria are not a fundamental part of the general metric definition.

src/metrics/definitions.md
jmccrosky: Discussed in thread above.

src/metrics/definitions.md
jmccrosky: Done.

src/metrics/definitions.md
jmccrosky: As part of the above definition, I've completely changed to metric definition.  It probably needs further work though, so I'm open to suggestions.

jmccrosky:
Thanks all!  I would like to flag that I would like to merge this on Wednesday so that it is easier to review for the broader MSG community - those that are unlikely to review content inside github are probably those from which we most need review.

I recognize that this will be merged "incomplete", but don't feel that there is much risk of harm from this.

src/metrics/definitions.md
wlach: I think that could work, as long as the definitions in-situ are kept short. I assume at least the short definitions are unlikely to change.

src/SUMMARY.md
wlach: > Regarding having the policy docs elsewhere, I'm not really opposed, but feel that it's nice to have an identified home for MSG and this seems like it will be it. It seems natural that people might expect to find documents related to our policies here - a nice parallel is the docs on contributing to DTMO that are linked at the bottom. But I don't feel that strongly. Any particular reason not to have the policy docs here?

I do worry a little bit about diluting docs.telemetry.mozilla.org's purpose / focus, as has happened in the past. But I can see the argument that this is the simplest place to put them for now.

src/metrics/definitions.md
wlach: Hmm, I guess I don't really see the point of a document that's just an abbreviated version of another? 

I think it would make more sense to provide some introductory content here on what you're trying to accomplish with the different types of detailed definitions in the child documents, then link to them explicitly here (as sort of a meta table of contents, [like the bigquery cookbook now does](https://docs.telemetry.mozilla.org/cookbooks/bigquery.html)). IMO some kind of story needs to be told about why people should read this section closely and carefully -- it doesn't need to be long (in fact it's probably better if it isn't).

jmccrosky: That's a fair point.  I'll make a pass of making this page more useful and am open to eventually removing it if we prefer to just promote its children one level.

jmccrosky:
OK, I'm merging this now to allow wider review.  There are of course further discussion and revisions needed and I'll make sure everyone is looped in on the new PR to clean this up.  I expect to start it within one week.

src/SUMMARY.md
harterrt: Adding some late comments for the record. These aren't pressing; no need to address these points.

> Regarding having the policy docs elsewhere, I'm not really opposed, but feel that it's nice to have an identified home for MSG and this seems like it will be it. 

I agree that we should have a home for MSG, but I don't think it should be DTMO. I share @wlach's concern about diluting DTMO. 

>  a nice parallel is the docs on contributing to DTMO that are linked at the bottom.

Good point. The `meta` docs are probably too extensive in DTMO. We should probably prefer to cover the contributing guidelines in the repository instead of in the rendered documentation.

PR #480, opened 2020-05-20
Fix broken internal links.

felixlawrence:
All these links were broken - presumably some content was moved or a file was renamed.


felixlawrence:
@wlach I just noticed that I might have caught you mid-refactor! Feel free to toss this PR and incorporate the fix into your ongoing work, if that's more convenient

PR #485, opened 2020-05-29
Add links to bit pattern docs

jklukas:
Follow-up to #462


jklukas:
Going ahead and merging as this as it's a small fixup for an oversight in the previous PR.

PR #487, opened 2020-05-31
Analysis Gotcha: Cycle lengths gradually moved from 8 to 4 weeks.

mnoorenberghe:
This can affect analysis done by version number as a lower number of events are likely to happen in the shorter time period.

Feel free to improve the wording. I didn't call out each cycle length change in the gradual process but can if people don't think it'll be too noisy.


wlach:
Thank you!

PR #492, opened 2020-06-03
Mention segment columns in clients_last_seen.

felixlawrence:
Regular Users v3 and New or Resurrected v3 have been added to the ETL - document where to find these new columns.

### DIFF
--- a/src/concepts/segments.md
+++ b/src/concepts/segments.md
@@ -25,9 +25,13 @@ so that your communication is forwards compatible.
 
 ### Regular users v3
 
+`clients_last_seen.is_regular_user_v3`
+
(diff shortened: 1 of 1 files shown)


src/concepts/segments.md
```diff
@@ -25,10 +25,14 @@ so that your communication is forwards compatible.
 
 ### Regular users v3
 
+`clients_last_seen.is_regular_user_v3`
```
jmccrosky, line 28: I don't feel too strongly, but it's possibly non-obvious that this is <tablename>.<fieldname>  Could spell it out "This segment is available in the X field of the Y table" if you like.

felixlawrence: I think that as we add more segments, the extra words might become tiresome? So I'll leave it for now

PR #494, opened 2020-06-05
Bug 1643341 - Add a section about "pings from robots" to analysis gotchas

wlach:


wlach:
Made a minor adjustment to make it clear this is for desktop. Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1644133 to handle the mobile case.

PR #499, opened 2020-06-10
Bug 1633452 - Merge our two query optimization guides into one

wlach:


src/cookbooks/bigquery/optimization.md
```diff
@@ -1,28 +1,165 @@
...
-    - See [_Writing Queries_](./querying.md#writing-queries) for examples.
-- Reduce data before using a JOIN
-    - Trim the data as early in the query as possible, before the query performs a JOIN. If you reduce data early in the processing cycle, shuffling and other complex operations only execute on the data that you need.
-    - Use sub queries with filters or intermediate tables or views as a way of decreasing sides of a join, prior to the join itself.
+# Optimizing BigQuery Queries
+
+When writing a query using [STMO](https://sql.telemetry.mozilla.org) or the BigQuery console, you can improve performance and reduce costs by learning how data is stored, how databases function, and what you can change about a query to take advantage of the storage structure and the data function.
+
+[Queries are charged by data scanned at \$5 per terabyte (TB)](https://cloud.google.com/bigquery/pricing#on_demand_pricing) so each 200 gigabytes of data scanned will cost \$1: on tables with hundreds of TBs of data (like the [main ping table](../../datasets/pings.md#main-ping) or [`clients_daily`](../../datasets/batch_view/clients_daily/reference.md)), costs **can add up very quickly**.
```
fbertsch, line 5: I don't want people to necessarily conflate reducing costs with improving performance. They are related but not the same. Some of the optimizations below will decrease costs, some won't. We could consider marking everything below that decreases costs, and note that here; or re-order it such that those are first.

src/cookbooks/bigquery/optimization.md
```diff
@@ -1,28 +1,165 @@
...
+  population (without sampling).
+  - Tables that include a `sample_id` field will usually have that as one of the clustering fields and you can efficiently scan random samples of users by specifying `WHERE sample_id = 0` (1% sample), `WHERE sample_id < 10` (10% sample), etc. This can be especially helpful with `main_summary`, `clients_daily`, and `clients_last_seen` which are very large tables and are all clustered on `sample_id`.
+- Many datasets also cluster on `normalized_channel`, corresponding to the channel of the product. If you are working with data that has different channels (for example, Firefox desktop), limit your initial query to a channel with a limited population like Nightly (in the case of Firefox desktop, do this by adding `WHERE normalized_channel='nightly'` to your query)
+- Select only the columns that you want (**Don't** use `SELECT *`)
+  - If you are experimenting with data or exploring data, use one of the [data preview options](https://cloud.google.com/bigquery/docs/best-practices-costs#preview-data) instead of `SELECT *`.
+- Use approximate algorithms: e.g., `approx_distinct(...)` instead of `COUNT(DISTINCT ...)`
```
fbertsch, line 20:
```suggestion
- Use [approximate algorithms](https://cloud.google.com/bigquery/docs/reference/standard-sql/approximate_aggregate_functions): e.g., `approx_count_distinct(...)` instead of `COUNT(DISTINCT ...)`
```

fbertsch: oops, see you have the link just below this

src/cookbooks/bigquery/optimization.md
```diff
@@ -1,28 +1,165 @@
...
-- Reference the data size prediction ("This query will process X bytes") in STMO and the BigQuery UI to help gauge the efficiency of your queries. You should see this number go down as you limit the range of `submission_date`s or include fewer fields in your `SELECT` statement. For clustered tables, this estimate won't take into account benefits from `LIMIT`s and `WHERE` clauses on clustering fields, so you'll need to compare to the actual "Data Scanned" after the query is run. [Queries are charged by data scanned at $5/TB](https://cloud.google.com/bigquery/pricing#on_demand_pricing) so each 200 GB of data scanned will cost $1; it can be useful to keep the data estimate below 200 GB while developing and testing a query to limit cost and query time, then open up to the full range of data you need when you have confidence in the results.
+  - WITH clauses are used primarily for readability because they are not materialized. For example, placing all your queries in WITH clauses and then running UNION ALL is a misuse of the WITH clause. If a query appears in more than one WITH clause, it executes in each clause.
```
fbertsch, line 26: This example of a misuse is confusing to me. In general it's fine to use CTEs any way users wish, just know they are not materialized.

src/cookbooks/bigquery/optimization.md
```diff
@@ -1,28 +1,165 @@
...
+
+```
+SELECT name
+FROM people
+```
+
+In this case, only the "name" row needs to be read. All the other lines of the
+file can be skipped.
+
+#### Data partitions
```
fbertsch, line 145: We should add a section here that describes clustering/sorting.

src/cookbooks/bigquery/optimization.md
```diff
@@ -1,28 +1,165 @@
...
+```
+SELECT *
+FROM people
+WHERE country == 'US'
+```
+
+In this case, the query engine no longer even has to read the file. It could just look at the path and realize that there is nothing of interest.
+
+Tables are usually partitioned based on dates; e.g., `submission_date` or `DATE(submission_timestamp)`.
+
```
fbertsch, line 161:
```suggestion

#### Data Ordering (clustering)

Another way to improve query performance is to select a subset of data on a field that the data is ordered by. In BigQuery, this is called "clustering". A clustered field is one which is sorted in the underlying data.

For example, if we wanted to get all of ages greater than age 40 in our table above, we might query like this:
`SELECT age FROM people WHERE age > 45`

Then we would scan all of the `age` field, starting from `27`, then `45`, then `5`.

However, if we sort the data on that field, our table would look like this:
\```
name,"Cadence","Ted","Emmanuel"
age,5,27,45
height,3.5,6.0,5.9
\```

Since data is stored on different files, we could ignore any `age` files that don't have data less than 40. So if Cadence and Ted's ages were in one file, and Emmanuel's in the next, we could skip reading that first file entirely. In that way, we can sometimes drastically reduce the amount of data we're reading.
```

src/cookbooks/bigquery/optimization.md
```diff
@@ -1,28 +1,165 @@
...
+#### Key takeaways
 
-You can locate a complete list of optimizations [here](https://cloud.google.com/bigquery/docs/best-practices-performance-overview) and cost optimizations [here](https://cloud.google.com/bigquery/docs/best-practices-costs)
+- Limit queries to a few columns that you need to reduce the volume of data that must be read
+- Filter the partitions to reduce the volume of data that you need
```
fbertsch, line 165:
```suggestion
- Filter the partitions to reduce the volume of data that you need
- Filter on clustered fields
```

src/cookbooks/bigquery/optimization.md
```diff
@@ -1,28 +1,165 @@
...
+more overhead.
+
+#### Columnar Stores
+
+Columnar turns the data sideways. For example, you can make a columnar version of the above data
+and still store it in CSV format:
+
+```
+name,"Ted","Emmanuel","Cadence"
+age,27,45,5
```
fbertsch, line 120: Side note: my daughter, Cadence, was 5 when I wrote this, she just turned 8 yesterday 😭

src/cookbooks/bigquery/optimization.md
wlach: Good point, I guess the point to emphasize is that reducing costs entails reducing the amount of data scanned. I can take a stab at adjusting this text to send that message. We can make further changes later if desired.

src/cookbooks/bigquery/optimization.md
wlach: I think I like this way better, less text. I'll apply this suggestion and delete the text below.

src/cookbooks/bigquery/optimization.md
wlach: How about this?

```suggestion
  - WITH clauses are used primarily for readability because they are not materialized: if a query appears in more than one WITH clause, it executes in each clause. Do not rely on them to optimize your query!
```

src/cookbooks/bigquery/optimization.md
wlach: LGTM, thanks!

src/cookbooks/bigquery/optimization.md
fbertsch: Perfect!

src/cookbooks/bigquery/optimization.md
wlach: Just updated the article further, adding some text to the introduction and doing the breakdown you suggested. I think this is better! Lmk what you think (when you have time)

src/cookbooks/bigquery/optimization.md
```diff
@@ -1,28 +1,189 @@
...
+
+Another way to improve query performance is to select a subset of data on a field that the data is ordered by. In BigQuery, this is called "clustering". A clustered field is one which is sorted in the underlying data.
+
+For example, if we wanted to get all of ages greater than age 40 in our table above, we might query like this:
+`SELECT age FROM people WHERE age > 45`
+
+Then we would scan all of the `age` field, starting from `27`, then `45`, then `5`.
+
+However, if we sort the data on that field, our table would look like this:
+\```
```
fbertsch, line 178: I can't remove these with a suggestion, but I had an extra `\` in here (and in the closing backticks).

elkevorh: Please revise as follows:
Then all `age` field are scanned, starting from `27`, proceeding to `45`, and ending with `5`.

If data is sorted based on that field, the table looks like this:

src/cookbooks/bigquery/optimization.md
elkevorh: I would change the following from:
When writing a query
To
When you write a query

src/cookbooks/bigquery/optimization.md
elkevorh: I would change this from:

Many datasets also cluster on `normalized_channel`, corresponding to the channel of the product. If you are working with data that has different channels (for example, Firefox desktop), limit your initial query to a channel with a limited population like Nightly (in the case of Firefox desktop, do this by adding `WHERE 

To 

Many datasets also cluster on `normalized_channel` and correspond to the channel of the product. If you work with data that has been assigned to different channels (for example, Firefox desktop), limit your initial query to a channel that has a limited population, such as Nightly (in the case of Firefox desktop, adding `WHERE . . . .


Change this from:

If you are experimenting with data or exploring data, use . . . 

To:

If you experiment with or explore data, use . . .

src/cookbooks/bigquery/optimization.md
elkevorh: Change fro:
If the SQL aggregation function you're using has an equivalent approximation function, the approximation function will yield faster query performance. 

To

If the SQL aggregation function that you use provides an equivalent approximation function, the approximation function then results in a much faster query performance. 
============

From

Reference the data size prediction ("This query will process X bytes") in STMO and the BigQuery UI to help gauge the efficiency of your queries. You should see this number go down as you limit the range of `submission_date`s or include fewer fields in your `SELECT` statement. For clustered tables, this estimate won't take into account benefits from `LIMIT`s and `WHERE` clauses on clustering fields, so you'll need to compare to the actual "Data Scanned" after the query is run. [Queries are charged by data scanned at $5/TB](https://cloud.google.com/bigquery/pricing#on_demand_pricing) so each 200 GB of data scanned will cost $1; it can be useful to keep the data estimate below 200 GB while developing and testing a query to limit cost and query time, then open up to the full range of data you need when you have confidence in the results.

To:


Reference the data size prediction ("This query processes X bytes") in STMO and the BigQuery UI to help evaluate the efficiency of your queries. You can view this number being decreased as you limit the range of `submission_date`s or include fewer fields in the `SELECT` statement. For clustered tables, this estimate does not take into account benefits from `LIMIT`s and `WHERE` clauses on clustering fields. Therefore, you need to compare to the actual "Data Scanned" after the query is run. [Queries are charged by data scanned at $5/TB](https://cloud.google.com/bigquery/pricing#on_demand_pricing). Each 200 GB of data scanned costs $1. You may want to keep the data estimate under 200 GB while you develop and test a query to limit cost and query time. You can then open it to the full range of data that you need when you have confidence in the results.

src/cookbooks/bigquery/optimization.md
elkevorh: From:

For example, if we wanted to get all of ages greater than age 40 in our table above, we might query like this:

To:

For example, if you want to retrieve information in the table above-listed table about all age groups with members who are older than 40 years, you may wan to query as follows:

elkevorh: I am sorry that there are typos in the above recommendation. Here are the fixes:
For example, if you want to retrieve information in the above-listed table about all age groups with members who are older than 40 years, you may want to query as follows:

elkevorh: From:
For example, if we wanted to get all of ages greater than age 40 in our table above, we might query like this:

To:
For example, if you want to retrieve information in the above-listed table about all age groups with members who are older than 40 years, you may want to query as follows:

elkevorh: From:
Then we would scan all of the `age` field, 

To:
Then you can scan all of the `age` field,

elkevorh: From:
However, if we sort the data on that field, our table would look like this:

To:
If you sort data on that field, the table then looks as follows:

elkevorh: From:
Since data is stored on different files, we could ignore any `age` files that don't have data less than 40. So if Cadence and Ted's ages were in one file, and Emmanuel's in the next, we could skip reading that first file entirely. In that way, we can sometimes drastically reduce the amount of data we're reading.

To:
You can ignore any `age` files that do not include any age-related data lower than 40 because data is stored in different files. If Cadence's and Ted's ages are included in one file and Emmanuel's age is included in the next file, you can then skip reading the first file. Consequently, the volume of data that is read in that manner can be drastically reduced.

elkevorh: Side note from Elke: The word 'since' should only be used when indicating a particular point in time. For example, "Since I graduated from college, I have been working as a technical writer."

Most of the time 'since' is incorrectly used and should be replaced with the word 'because'. In other words, "As a cause of this and that occurring, I am no longer a technical writer." Many times we use in our verbal communication 'since' in this situation even though we should have said something like this "I am no longer a technical writer because (not since) I made so many mistakes. However, I stopped working since I got engaged (a particular point in time).

limpaopao:
> 🚢

PR #501, opened 2020-06-12
Initial proposal for DAU definition.

jmccrosky:
This is perhaps not complete, but wanted to get some initial opinions.

### DIFF
--- a/src/metrics/metrics.md
+++ b/src/metrics/metrics.md
@@ -1,6 +1,6 @@
 # Metrics
 
-Here we provide definitions and information about our standard metrics.
+Here we provide definitions and information about our standard metrics.  We use some terminology that is defined [here](https://docs.telemetry.mozilla.org/concepts/terminology.html).
 
 ## Table of Contents
 
@@ -22,7 +22,7 @@ The concept of being "active" is defined as in the [`usage criterion`](./usage.m
 
 ### Caveats
 
-If the number of users stays constant, but the average number of active profiles per user increases, this metric will tend to increase.
+If the number of users stays constant, but the average number of active profiles per user increases, this metric will tend to increase.  For more details on the relationship between users and profiles, see [here](https://docs.telemetry.mozilla.org/concepts/profile/realworldusage.html).
 
 ### Dashboards
 
@@ -50,6 +50,8 @@ WHERE
 GROUP BY
   date
 ```
+
+You can run this query on [STMO](https://sql.telemetry.mozilla.org/queries/72012/source).
 
 -----
 
--- a/src/metrics/metrics.md
+++ b/src/metrics/metrics.md
@@ -18,9 +18,9 @@ Daily Active Users or DAU counts the number of unique profiles active in the pro
 
 DAU counts unique profiles. Keep in mind that a [profile is not necessarily a user](/concepts/analysis_gotchas.html#profiles-vs-users)
 
-The precise concept of active varies by product, but generally, active users are defined as unique profiles who've sent a `main` ping (on desktop) or a `baseline` ping (on mobile).  The precise criteria are defined in the [`usage criterion`](./usage.md) section of this documentation.
+The standard concept of active varies by product, but generally, active users are defined as unique profiles who've sent a `main` ping (on desktop) or a `baseline` ping (on mobile).  The precise criteria are defined in the [`usage criterion`](./usage.md) section of this documentation.
 
-We can also occasionally restrict the metric to alternative usage critera. It's **critical to clearly state any non-standard usage criteria on the metric**. The metrics team suggest the following format: `DAU(usage criteria)`. For example, we might be interested in the count of users who've viewed more than 5 URIs in a day. We'd denote that metric as `DAU(URI > 5)`. 
+We can also occasionally restrict the metric to alternative usage critera. It's **critical to clearly state any non-standard usage criteria on the metric**. The metrics team suggest the following format: `DAU(usage criteria)`. For example, we might be interested in the count of users who've viewed more than 5 URIs in a day. We'd denote that metric as `DAU(URI > 5)`.
 
 We also some common alternative usage criteria documented in [`usage criterion`](./usage.md).
 
--- a/src/metrics/metrics.md
+++ b/src/metrics/metrics.md
@@ -22,7 +22,7 @@ The concept of being "active" is defined as in the [`usage criterion`](./usage.m
 
 ### Caveats
 
-If the number of users stays constant, but the average number of active profiles per user increases, this metric will tend to increase.  For more details on the relationship between users and profiles, see [here](https://docs.telemetry.mozilla.org/concepts/profile/realworldusage.html).
+If the number of users stays constant, but the average number of active profiles per user increases, this metric will tend to increase.  For more details on the relationship between users and profiles, see [here](https://docs.telemetry.mozilla.org/concepts/analysis_gotchas.html#profiles-vs-users).
 
 ### Dashboards
 
--- a/src/metrics/metrics.md
+++ b/src/metrics/metrics.md
@@ -34,25 +34,20 @@ This metric is available on the [standard Growth and Usage Dashboard (GUD)](http
 
 ### Tables
 
-DAU can easily be calculated from the Smoot tables; for example:
+DAU can easily be calculated from the [Exact MAU tables](https://docs.telemetry.mozilla.org/datasets/bigquery/exact_mau/reference.html); for example:
 
 ```sql
 SELECT
-  date,
+  submission_date,
   SUM(dau) AS dau
 FROM
-  `moz-fx-data-shared-prod.telemetry.smoot_usage_day_0`
--- You must always filter to a single usage criterion
--- For a full list of available usage criteria, see https://sql.telemetry.mozilla.org/queries/65338/source
+  `moz-fx-data-derived-datasets.telemetry.firefox_desktop_exact_mau28_by_dimensions_v1`
 WHERE
-  usage = 'Any Firefox Desktop Activity'
-
   -- You define your slice using additional filters here.
-  -- Available dimensions: app_name, app_version, country, locale, os, os_version, channel
-  AND country IN ('US', 'GB', 'CA', 'FR', 'DE')
+  country IN ('US', 'GB', 'CA', 'FR', 'DE')
 
 GROUP BY
-  date
+  submission_date
 ```
 
 You can run this query on [STMO](https://sql.telemetry.mozilla.org/queries/72012/source).
--- a/src/metrics/metrics.md
+++ b/src/metrics/metrics.md
@@ -22,7 +22,7 @@ The concept of being "active" is defined as in the [`usage criterion`](./usage.m
 
 ### Caveats
 
-If the number of users stays constant, but the average number of active profiles per user increases, this metric will tend to increase.  For more details on the relationship between users and profiles, see [here](https://docs.telemetry.mozilla.org/concepts/analysis_gotchas.html#profiles-vs-users).
+If the number of users stays constant, but the average number of active profiles per user increases, this metric will tend to increase.  For more details on the relationship between users and profiles, see [the profiles vs users section in analysis gotchas](https://docs.telemetry.mozilla.org/concepts/analysis_gotchas.html#profiles-vs-users).
 
 ### Dashboards
 
--- a/src/metrics/metrics.md
+++ b/src/metrics/metrics.md
@@ -16,7 +16,7 @@ Daily Active Users or DAU counts the number of unique profiles active in the pro
 
 ### Details
 
-The concept of a "unique profile" may vary from product to product and the relationship between that concept and actual users may also vary.  As such care should be taken interpreting this metric as an actual count of users.
+DAU counts unique profiles. Keep in mind that a [profile is not necessarily a user](/concepts/analysis_gotchas.html#profiles-vs-users)
 
 The concept of being "active" is defined as in the [`usage criterion`](./usage.md) section and will vary by product.  A product may even have multiple possible `usage criteria`, but there will be one default choice that is our best approximation of general product usage.
 
--- a/src/metrics/metrics.md
+++ b/src/metrics/metrics.md
@@ -2,9 +2,58 @@
 
 Here we provide definitions and information about our standard metrics.
 
-Table of contents with links to all metrics.
+## Table of Contents
+
+<!-- toc -->
 
 -----
+
+## Daily Active Users (DAU)
+
+### Overview
+
+DAU counts the number of unique profiles active in the product on each day.  This is intended to approximate the number of people using the product each day.
+
+### Details
+
+The concept of a "unique profile" may vary from product to product and the relationship between that concept and actual users may also vary.  As such care should be taken interpreting this metric as an actual count of users.
+
+The concept of being "active" is defined as in the [`usage criterion`](./usage.md) section and will vary by product.  A product may even have multiple possible `usage criteria`, but there will be one default choice that is our best approximation of general product usage.
+
+### Caveats
+
+If the number of users stays constant, but the average number of active profiles per user increases, this metric will tend to increase.
+
+### Dashboards
+
+This metric is available on the [standard Growth and Usage Dashboard (GUD)](https://go.corp.mozilla.com/gud) for most products and with some slicing available.
+
+### Tables
+
+DAU can easily be calculated from the Smoot tables; for example:
+
+```
+SELECT
+  date,
+  SUM(dau) AS dau
+FROM
+  `moz-fx-data-shared-prod.telemetry.smoot_usage_day_0`
+-- You must always filter to a single usage criterion
+-- For a full list of available usage criteria, see https://sql.telemetry.mozilla.org/queries/65338/source
+WHERE
+  usage = 'Any Firefox Desktop Activity'
+
+  -- You define your slice using additional filters here.
+  -- Available dimensions: app_name, app_version, country, locale, os, os_version, channel
+  AND country IN ('US', 'GB', 'CA', 'FR', 'DE')
+
+GROUP BY
+  date
+```
+
+-----
+
+Reference below.
 
 {{Metric name}}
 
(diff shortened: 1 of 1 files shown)


src/metrics/metrics.md
```diff
@@ -2,10 +2,59 @@
...
+
+<!-- toc -->
 
 -----
 
+## Daily Active Users (DAU)
+
+### Overview
+
+DAU counts the number of unique profiles active in the product on each day.  This is intended to approximate the number of people using the product each day.
```
wlach, line 15:
```suggestion
Daily Active Users or DAU counts the number of unique profiles active in the product on each day.  This is intended to approximate the number of people using the product each day.
```

src/metrics/metrics.md
```diff
@@ -2,10 +2,59 @@
...
+
+### Details
+
+The concept of a "unique profile" may vary from product to product and the relationship between that concept and actual users may also vary.  As such care should be taken interpreting this metric as an actual count of users.
+
+The concept of being "active" is defined as in the [`usage criterion`](./usage.md) section and will vary by product.  A product may even have multiple possible `usage criteria`, but there will be one default choice that is our best approximation of general product usage.
+
+### Caveats
+
+If the number of users stays constant, but the average number of active profiles per user increases, this metric will tend to increase.
```
wlach, line 25: I think we should go into more detail about this... somewhere.

There's this bit of good explanation on the distinction between pings and users here: https://docs.telemetry.mozilla.org/concepts/analysis_intro.html#caveats but it's somewhat incomplete and mixed in with discussion of an obsolete system (telemetry aggregates).

I can file a bug for a followup.

src/metrics/metrics.md
```diff
@@ -2,10 +2,59 @@
...
+WHERE
+  usage = 'Any Firefox Desktop Activity'
+
+  -- You define your slice using additional filters here.
+  -- Available dimensions: app_name, app_version, country, locale, os, os_version, channel
+  AND country IN ('US', 'GB', 'CA', 'FR', 'DE')
+
+GROUP BY
+  date
+```
```
wlach, line 52: Can you make this into a real query and link to it on STMO?

src/metrics/metrics.md
```diff
@@ -2,10 +2,59 @@
...
+
+### Dashboards
+
+This metric is available on the [standard Growth and Usage Dashboard (GUD)](https://go.corp.mozilla.com/gud) for most products and with some slicing available.
+
+### Tables
+
+DAU can easily be calculated from the Smoot tables; for example:
+
+```
```
wlach, line 35:
```suggestion
```sql
```

src/metrics/metrics.md
wlach, line 15: We should define with a profile is in terminology (https://docs.telemetry.mozilla.org/concepts/terminology.html)

src/metrics/metrics.md
```diff
@@ -2,10 +2,59 @@
 
 Here we provide definitions and information about our standard metrics.
```
wlach, line 3: Should we link to the terminology section (https://docs.telemetry.mozilla.org/concepts/terminology.html) in this introduction? I think some of the material there may be helpful in interpreting what's in DAU.

jmccrosky: Sure, adding.  I think that section needs a lot more material - might be something we can work on as part of this initiative - but it's still useful to reference.

src/metrics/metrics.md
jmccrosky: Maybe this is a better thing to reference?  It probably needs to be updated, but isn't bad.  I'll put it in for now, but feel free to debate ;)

src/metrics/metrics.md
jmccrosky: added

src/metrics/metrics.md
wlach: Checking again, we have this which is actually reasonably up to date: https://docs.telemetry.mozilla.org/concepts/analysis_gotchas.html#profiles-vs-users

Are you ok with referencing that?

jmccrosky: yup; updated!

src/metrics/metrics.md
```diff
@@ -1,11 +1,62 @@
...
+
+### Details
+
+The concept of a "unique profile" may vary from product to product and the relationship between that concept and actual users may also vary.  As such care should be taken interpreting this metric as an actual count of users.
+
+The concept of being "active" is defined as in the [`usage criterion`](./usage.md) section and will vary by product.  A product may even have multiple possible `usage criteria`, but there will be one default choice that is our best approximation of general product usage.
+
+### Caveats
+
+If the number of users stays constant, but the average number of active profiles per user increases, this metric will tend to increase.  For more details on the relationship between users and profiles, see [here](https://docs.telemetry.mozilla.org/concepts/analysis_gotchas.html#profiles-vs-users).
```
wlach, line 25:
```suggestion
If the number of users stays constant, but the average number of active profiles per user increases, this metric will tend to increase.  For more details on the relationship between users and profiles, see [the profiles vs users section in analysis gotchas](https://docs.telemetry.mozilla.org/concepts/analysis_gotchas.html#profiles-vs-users).
```

src/metrics/metrics.md
```diff
@@ -1,11 +1,62 @@
...
+
+### Overview
+
+Daily Active Users or DAU counts the number of unique profiles active in the product on each day.  This is intended to approximate the number of people using the product each day.
+
+### Details
+
+The concept of a "unique profile" may vary from product to product and the relationship between that concept and actual users may also vary.  As such care should be taken interpreting this metric as an actual count of users.
+
+The concept of being "active" is defined as in the [`usage criterion`](./usage.md) section and will vary by product.  A product may even have multiple possible `usage criteria`, but there will be one default choice that is our best approximation of general product usage.
```
felixlawrence, line 21: I think it's a bad idea to include usage criteria here. I think that the industry standard is that "DAU" is a concrete metric, with "Active" having some fixed definition that's relevant to the product.

I think we're asking for trouble if we define "DAU" as a parametrised metric. Most Mozilla staff are going to talk about "Desktop DAU" rather than "Desktop DAU with the usage criterion of sending a ping". This would mean that their usage of the terminology would be technically incorrect, and underspecified.

For what benefit? So that we could say "Desktop DAU with the usage criterion of having nonzero active hours", instead of "The number of desktop clients per day with nonzero active hours"? Is there any advantage to bringing the "DAU" acronym in here and diluting it?


I understand that in the context of building GUD, it made sense to write one query/function that solves a class of problems, rather than many queries/functions that each solve a narrow problem. In the context of writing a tool, or designing a UI, you want a small set of things that can combine to solve a large class of problems.

But here in the context of defining key metrics, I claim that we don't want to define broad classes of metrics: we want to define the key metrics that people are watching and making decisions from. And we don't want to leave room for ambiguity

src/metrics/metrics.md
```diff
@@ -1,11 +1,62 @@
...
+
+## Daily Active Users (DAU)
+
+### Overview
+
+Daily Active Users or DAU counts the number of unique profiles active in the product on each day.  This is intended to approximate the number of people using the product each day.
+
+### Details
+
+The concept of a "unique profile" may vary from product to product and the relationship between that concept and actual users may also vary.  As such care should be taken interpreting this metric as an actual count of users.
```
felixlawrence, line 19: What is the reader supposed to take away from this paragraph? "Don't think that you'll understand what this metric is by the time you've finished reading the document - you're going to have to hunt elsewhere to get unspecified extra knowledge"? Can we do some of the legwork for the reader and be more useful?

limpaopao:
发自我的iPhone

src/metrics/metrics.md
harterrt: Perhaps we can just link to the "profile vs users" section as we did below. 

Suggested revision:

```suggestion
DAU counts unique profiles. Keep in mind that a [profile is not necessarily a user](/concepts/analysis_gotchas.html#profiles-vs-users)
```

src/metrics/metrics.md
jmccrosky: GUD has become the standard tool for measuring DAU and explicitly uses usage criteria.  I agree that we want a "standard DAU" per product, but allowing DAU to be in terms of alternative usage criteria doesn't seem to be asking too much of our users.

I tried to make this clear in the text "there will be one default choice that is our best approximation of general product usage" and that if we just say Desktop DAU, that is well-defined.  Do you think the text isn't clear enough on this point?

src/metrics/metrics.md
```diff
@@ -1,11 +1,62 @@
 # Metrics
 
-Here we provide definitions and information about our standard metrics.
+Here we provide definitions and information about our standard metrics.  We use some terminology that is defined [here](https://docs.telemetry.mozilla.org/concepts/terminology.html).
```
wlach, line 3: A few extra things jumped out at me reading this over again:

* Best practice is not to use "we" or "our" in technical documentation (it's not really clear "who" that means, especially in material produced by so many people)
* In general try not to use "here" when linking to things, just talk about it and linkify the concept: this is both more accessible and easier for the reader -- https://medium.com/@heyoka/dont-use-click-here-f32f445d1021 

```suggestion
This section contains definitions and information about standard metrics used at Mozilla.  You may wish to refer to the [terminology section](../concepts/terminology.html) while reading this document, in case a particular concept is not clear.
```

src/metrics/metrics.md
felixlawrence: > allowing DAU to be in terms of alternative usage criteria doesn't seem to be asking too much of our users.

It's asking too much of me! It means that I need to say "DAU for sends a ping" instead of "DAU", because the latter would be undefined without a usage criterion.

> I tried to make this clear in the text "there will be one default choice that is our best approximation of general product usage" and that if we just say Desktop DAU, that is well-defined. Do you think the text isn't clear enough on this point?

Yes, I think it's unclear: nowhere in this document does it specify what that default choice is! Someone could carefully comb this document, which seeks to define DAU, and they would be unable to learn that Firefox Desktop DAU is the number of profiles for which a ping is received on a given day!

And to be clear, I think the best solution for this document is to define "DAU" with a single usage criterion per platform, rather than to parametrise it and specify that default parameter somewhere in the document.

jmccrosky: I think we're thinking along the same lines and maybe just need to make the text better reflect what we're thinking.

"DAU" should mean a particular thing per product.  But that thing may not always be "sent a ping".  We already have some confusion around "which ping" for Fenix and, although this is not the place to discuss it, I hope we move to requiring non-zero active hours or some similar restriction for at least desktop in the future.

We don't want to have to have a separate DAU metric defined for each product.  Which is why we can have a single parameterized definition, while making clear that there is a standard parameter for each platform.

So maybe the text should make clear that DAU counts active users of a product and how we define active use is in the usage criteria section.

Sound fair?

felixlawrence: I think this definition is not complete unless it includes the usage criteria for each platform, on this page. Otherwise this page does not give a complete answer to the basic question "what is Firefox Desktop DAU?"! Shouldn't this page answer that question without sending people on an expedition?

jmccrosky: I guess the only point of disagreement now is whether to list all the standard usage criteria with the metric description.  If we do so, we're repeating the list of criteria on many metric descriptions, as they're needed to DAU. MAU, WAU, Intensity, probably Retention, and many others.  Doesn't it seem cleaner to have a separate list of usage criteria?  Asking a user to look at two pages to thoroughly understand a metric doesn't seem terrible.

src/metrics/metrics.md
```diff
@@ -1,11 +1,62 @@
...
+
+If the number of users stays constant, but the average number of active profiles per user increases, this metric will tend to increase.  For more details on the relationship between users and profiles, see [the profiles vs users section in analysis gotchas](https://docs.telemetry.mozilla.org/concepts/analysis_gotchas.html#profiles-vs-users).
+
+### Dashboards
+
+This metric is available on the [standard Growth and Usage Dashboard (GUD)](https://go.corp.mozilla.com/gud) for most products and with some slicing available.
+
+### Tables
+
+DAU can easily be calculated from the Smoot tables; for example:
```
tdsmith, line 33: Where's the reference documentation for these?

src/metrics/metrics.md
harterrt: What was the resolution here?

src/metrics/metrics.md
harterrt, line 21: Here's a draft revision given the conversation in #msg-internal

```suggestion
By default, active users are users who've sent a `main` ping (on desktop) or a `baseline` ping (on mobile).

This is a pretty low bar to call a user "active", so we'll occasionally restrict the metric to users with more activity. It's **critical to clearly document any additional qualifications on the metric**. The metrics team suggest the following format: `DAU(qualification)`. For example, we might be interested in the count of users who've viewed more than 5 URIs in a day. We'd denote that metric as `DAU(URI > 5)`. 
```

wlach: In principle I'm fine with this phrasing. IMO it would be better to use the "profiles" terminology here, to reinforce that a unique profile (client) is not the same as a user. Something like:

`By default, "active users" are defined as unique profiles who have sent a `main` ping (on desktop) or a `baseline` ping (on mobile).`

src/metrics/metrics.md
jmccrosky: @jklukas do we have anything we can point to here?

src/metrics/metrics.md
```diff
@@ -1,11 +1,66 @@
...
+
+## Daily Active Users (DAU)
+
+### Overview
+
+Daily Active Users or DAU counts the number of unique profiles active in the product on each day.  This is intended to approximate the number of people using the product each day.
+
+### Details
+
+DAU counts unique profiles. Keep in mind that a [profile is not necessarily a user](/concepts/analysis_gotchas.html#profiles-vs-users)
```
jklukas, line 19:
```suggestion
DAU counts unique profiles. Keep in mind that a [profile is not necessarily a user](/concepts/analysis_gotchas.html#profiles-vs-users).
```

src/metrics/metrics.md
```diff
@@ -1,11 +1,66 @@
...
+
+The standard concept of active varies by product, but generally, active users are defined as unique profiles who've sent a `main` ping (on desktop) or a `baseline` ping (on mobile).  The precise criteria are defined in the [`usage criterion`](./usage.md) section of this documentation.
```
jklukas, line 21:
```suggestion
The standard concept of active varies by product, but generally, active users are defined as unique profiles that have sent a `main` ping (on desktop) or a `baseline` ping (on mobile).  The precise criteria are defined in the [`usage criterion`](./usage.md) section of this documentation.
```

To be consistent with the stance that profiles are not people.

src/metrics/metrics.md
```diff
@@ -1,11 +1,66 @@
...
+
+We can also occasionally restrict the metric to alternative usage critera. It's **critical to clearly state any non-standard usage criteria on the metric**. The metrics team suggest the following format: `DAU(usage criteria)`. For example, we might be interested in the count of users who've viewed more than 5 URIs in a day. We'd denote that metric as `DAU(URI > 5)`.
```
jklukas, line 23:
```suggestion
We can also restrict the metric to alternative usage critera. It's **critical to clearly state any non-standard usage criteria on the metric**. The metrics team suggest the following format: `DAU(usage criterion)`. For example, we might be interested in the count of profiles that have viewed more than 5 URIs in a day. We'd denote that metric as `DAU(URI > 5)`.
```

src/metrics/metrics.md
```diff
@@ -1,11 +1,66 @@
...
+
+We also some common alternative usage criteria documented in [`usage criterion`](./usage.md).
```
jklukas, line 25:
```suggestion
Some common alternative usage criteria documented in the [`usage criterion` section](./usage.md).
```

src/metrics/metrics.md
jklukas: We do not have reference documentation for these, and I do not consider the format of these tables stable enough to feel comfortable recommending that data users rely on them. The closest we have to reference documentation is these two STMO queries:

https://sql.telemetry.mozilla.org/queries/65338/source
https://sql.telemetry.mozilla.org/queries/65337/source

We have existing documentation in DTMO for DAU queries on top of the exact_mau28 tables and on top of clients_daily which are stable interfaces. I hope in H2 to refactor smoot tables and better codify the mapping of backend fields to logical application names and usage criteria, but there is currently risk in hitting the smoot tables.

jmccrosky: OK.  I'll use exact_mau28 instead and add the link.  Seems like it's better to stick with a standard and as high-level as possible of a table.  We may want to consider adding additional information for lower level queries at some point (like how to calculate MAU from raw data).   But perhaps that can wait.

jmccrosky:
Thanks @jklukas 

One thing that's raised is we may eventually want more example queries, to show how to calculate DAU for other products or for cases where we use a table that doesn't already have a dau column.  Will defer this though.

src/metrics/metrics.md
```diff
@@ -1,11 +1,61 @@
...
+
+### Dashboards
+
+This metric is available on the [standard Growth and Usage Dashboard (GUD)](https://go.corp.mozilla.com/gud) for most products and with some slicing available.
+
+### Tables
+
+DAU can easily be calculated from the [Exact MAU tables](https://docs.telemetry.mozilla.org/datasets/bigquery/exact_mau/reference.html); for example:
+
+```sql
```
felixlawrence, line 39: Similar to [this comment thread on the day 2-7 activation doc](https://docs.google.com/document/d/1HwlX6E3NzPkudVLBJXPcS49XpgwfxIIugXO0Y7kNlKM/edit?disco=AAAAGrjNdTQ), it would be good if there was a query that used a less-raw dataset - one that is more useful to data scientists wanting to do new things.

But I don't see this as a blocker for this PR, because I think DSes are going to either know how to do it, work out how to do it, or can just ask for help in less time than it would take to edit this document

src/metrics/metrics.md
tdsmith: Thank you both!

src/metrics/metrics.md
jmccrosky: Totally agree.  We discussed that above and I think having an example from both the highest-level and easiest table and also the rawest.  Will defer that for the next PR though.

PR #515, opened 2020-07-07
Add day 2-7 activation metric

wlach:


.spelling
```diff
@@ -113,6 +113,7 @@ filesystem
 firefox-data-docs
 FirefoxReality
 FireTV
+frac
```
wlach, line 116: This is to get around the fix that mdspell checks the spelling on MathJax blocks 🙄

PR #516, opened 2020-07-08
Use package.json for installing spell/link checking

wlach:
The primary motivation for this is that I suspect netlify
gets confused by repositories without a package.json file,
but I think this is also a better practice for installing
dependencies which are local to a particular project.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -190,6 +190,7 @@ nbconvert
 nbviewer
 nginx
 ni
+node.js
 non-e10s
 NoSQL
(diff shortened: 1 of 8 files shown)


wlach:
> The primary motivation for this is that I suspect netlify
> gets confused by repositories without a package.json file,
> but I think this is also a better practice for installing
> dependencies which are local to a particular project.

Turns out the issue was repository size, fixed by #517. This is still worth landing IMO.

PR #517, opened 2020-07-08
Optimize png files with pngcrush

wlach:
Makes the checked out repository a bit smaller


wlach:
I'm just going to land it, it's not controversial at all (just smaller file sizes). Hoping this might fix netlify issues?

PR #522, opened 2020-07-13
Segments rename and mau

felixlawrence:
* Use "user states" over "segments"
* Recommend use of "every segment MAU"

### DIFF
--- a/src/concepts/segments.md
+++ b/src/concepts/segments.md
@@ -70,7 +70,7 @@ User states are found as columns in the `clients_last_seen` dataset: the user st
 Users can move in or out of specific user states part way through a week or a month.
 This poses a conundrum if we want to plot the WAU or MAU for a user state.
 
-Our convention is to _count the number of distinct users who were active in user state in the period_.
+Our convention is to _count the number of distinct users who were active in the user state in the period_: e.g. "MAU(sent a ping as a regular user v3)".
 So if a user was active as a regular user v3 on e.g. the second day of a 28-day window, then they will contribute to "regular user v3 MAU" regardless of whether they lost their "regular user v3" status at any point in the 28-day window.
 
 Since many user states use the full extent of the `*_bits` column wizardry in `clients_last_seen`, you'll have to query WAU or MAU the old fashioned way:
--- a/src/concepts/segments.md
+++ b/src/concepts/segments.md
@@ -99,7 +99,7 @@ If you need to break MAU down into the sum of MAU for various user states, then 
 It might be tempting to assign users to whichever user state they happened to be in at the end of the window: this quantity is easy to query.
 But many of the user states were defined to be meaningful _on days the users were active_.
 "Regular users v3" is predictive of retention, _given that the user was active on the day of interest as a regular user_.
-A user can be "new or irregular v3" for only one day in a 28-day period: unless they appear for the first time on the last day of the month, the user will not qualify as "new or irregular v3" at the end of the MAU window!
+A user can be "new or resurrected v3" for only one day in a 28-day period: unless they appear for the first time on the last day of the month, the user will not qualify as "new or resurrected v3" at the end of the MAU window!
 So beware this trap and try to only use user states on days the users are active.
 
 
--- a/src/concepts/segments.md
+++ b/src/concepts/segments.md
@@ -101,6 +101,7 @@ If you need to break MAU down into the sum of MAU for various user states, then 
 It might be tempting to assign users to whichever user state they happened to be in at the end of the window: this quantity is easy to query.
 But many of the user states were defined to be meaningful _on days the users were active_.
 "Regular users v3" is predictive of retention, _given that the user was active on the day of interest as a regular user_.
+If someone has been using the browser every day for a year but then suddenly churns, then it's misleading to consider them to be active as a "not regular user v3" for MAU on the period ending the 15th day after their last activity.
 A user can be "new or resurrected v3" for only one day in a 28-day period: unless they appear for the first time on the last day of the month, the user will not qualify as "new or resurrected v3" at the end of the MAU window!
 So beware this trap and try to only use user states on days the users are active.
 
--- a/src/concepts/segments.md
+++ b/src/concepts/segments.md
@@ -1,74 +1,76 @@
-# Segments (beta)
+# User states/segments
 
-A segment is a group of clients who fit a set of criteria at a point in time. 
-The set of criteria itself can also be referred to as a "segment".
+A user state is a group of clients who fit a set of criteria at a point in time. 
+The set of criteria itself can also be referred to as a "user state".
 
-Typically you'll use segments to gain more insight into what is going on, by asking 
+In data science these user states would normally be called "segments"; we have been told to avoid that term to avoid confusion with marketing's segments.
+
+Typically you'll use user states to gain more insight into what is going on, by asking 
 "Regarding the thing I'm interested in, 
-do users in different segments behave differently, 
+do users in different user states behave differently, 
 and what insights does this give me into the users and the product?" 
 For example, "In this experiment, how do new users react to this feature, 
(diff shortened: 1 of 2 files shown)


src/concepts/segments.md
```diff
@@ -67,11 +67,40 @@ User states are found as columns in the `clients_last_seen` dataset: the user st
...
+    AND cls.submission_date BETWEEN '2020-04-01' AND '2020-07-01'
+WHERE cls.sample_id = 42
+    AND cls.is_regular_user_v3
+    AND cls.days_since_seen = 0
+GROUP BY dates.d
+ORDER BY dates.d
+```
+
+Our convention has the potentially counterintuitive consequence that a user can count towards "regular user v3 MAU" and "not regular user v3 MAU" for the same 28-day window.
+If you need to break MAU down into the sum of MAU for various user states, then in this instance you would need to break it down into "regular v3", "not regular v3", and "transitioning".
```
irrationalagent, line 97: It seems like transitioning essentially means "qualified as both regular and not regular in the same period but at different times". 

This is a nit, but to me, transitioning implies a one-time shift either into or out of a state, where in this case it could be a lot messier, e.g. dropping in and then out again multiple times. But I'm not sure I have a better term for "transitioning" other than "both regular and not-regular at different times". "Irregularly regular" is just awful. Maybe not-regular, full-time-regular, and part-time-regular?

src/concepts/segments.md
```diff
@@ -67,11 +67,40 @@ User states are found as columns in the `clients_last_seen` dataset: the user st
...
+
+It might be tempting to assign users to whichever user state they happened to be in at the end of the window: this quantity is easy to query.
+But many of the user states were defined to be meaningful _on days the users were active_.
+"Regular users v3" is predictive of retention, _given that the user was active on the day of interest as a regular user_.
+A user can be "new or irregular v3" for only one day in a 28-day period: unless they appear for the first time on the last day of the month, the user will not qualify as "new or irregular v3" at the end of the MAU window!
```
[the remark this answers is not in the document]
felixlawrence: Eek - yes, thanks for catching this!

src/concepts/segments.md
felixlawrence: Yeah, good call. "both" will do in the context of this sentence. It might not be a great option in practice - i.e. out of context on a graph legend - but it gets the point across for this document. So let's leave the search for a "great option in practice" as an exercise for the reader :grinning:

felixlawrence:
Now just waiting for @jmccrosky to look at this next week once he's back from PTO, to see if the MAU/WAU stuff is something he can live with.

src/concepts/segments.md
```diff
@@ -1,75 +1,106 @@
...
+    ON cls.submission_date BETWEEN DATE_SUB(dates.d, INTERVAL 27 DAY) AND dates.d
+    AND cls.submission_date BETWEEN '2020-04-01' AND '2020-07-01'
+WHERE cls.sample_id = 42
+    AND cls.is_regular_user_v3
+    AND cls.days_since_seen = 0
+GROUP BY dates.d
+ORDER BY dates.d
+```
+
+Our convention has the potentially counterintuitive consequence that a user can count towards "regular user v3 MAU" and "not regular user v3 MAU" for the same 28-day window.
```
jklukas, line 96: This sentence is where I see the terminology really break down. The idea of MAU for "not regular user v3" is problematic because we are marking users as "active" for a specific definition of _not_ being active.

This quantity may well be a useful one to calculate and observe, but I'd like to drop the MAU/WAU terminology at this point. This could be something like "monthly unique users in state"?

felixlawrence: > we are marking users as "active" for a specific definition of not being active

It's marking users as "active as not a regular user" (i.e. their browser submitted a ping on a day they didn't qualify to be a regular user), similar to how we'd mark users as "active not in the US" vs "active in the US". Do I need a better segment name than "not regular user" for "not regular user v3 MAU", or is it the concept itself the issue?

"monthly unique users in state" loses the implication that the users were active while being in that state - and it's important to keep that implication, otherwise for example regular users can never churn without appearing to transition through other states including "new or resurrected"(!)

src/concepts/segments.md
```diff
@@ -1,75 +1,106 @@
-# Segments (beta)
+# User states/segments
 
-A segment is a group of clients who fit a set of criteria at a point in time. 
-The set of criteria itself can also be referred to as a "segment".
+A user state is a group of clients who fit a set of criteria at a point in time. 
+The set of criteria itself can also be referred to as a "user state".
 
-Typically you'll use segments to gain more insight into what is going on, by asking 
+In data science these user states would normally be called "segments"; we have been told to avoid that term to avoid confusion with marketing's segments.
```
jmccrosky, line 6: We've already made clear that user states and segments are more or less synonymous.  I would argue that this sentence is unnecessary, potentially divisive, and can be removed.

src/concepts/segments.md
jmccrosky: I think that @jklukas makes a good suggestion about not referring to this quantity as MAU.  You can certainly use it, but having another name makes its strange behaviours less potentially-confusing.

I would still support the end-of-window definition that you describe below as the MAU definition.  Of your two arguments against it in the text below:

1) I'm sure that being a end-of-window regular user MAU is also predictive of retention and, in any case, predicting retention isn't the only reason to care about regular users
2) Being a new profile is something that happens on a discrete day.  Using MAU of a segment is not really a reasonable way to try to count discrete events.

felixlawrence: Further out-of-band communication with @jklukas resulted in the "MAU(sent a ping as a regular user v3)" language that's in the current version of this PR. The current version of the text tries to establish "regular user v3 MAU" as a shorthand for the verbose "MAU(sent a ping as a regular user v3)"

> I'm sure that being a end-of-window regular user MAU is also predictive of retention and, in any case, predicting retention isn't the only reason to care about regular users

It wasn't designed to be used in this way; if we want a quantity that works optimally for this then we should specifically design a quantity for this purpose, and it will likely be a different quantity.

src/concepts/segments.md
felixlawrence: This is the only place in the document where we explicitly say that "user states" and what DSes call "segments" are the same thing. I'll remove the reason we went with "user states" unless you can think of an innocuous phrasing.

src/concepts/segments.md
xluo-ds: Is it necessary that we apply user state onto WAU and MAU? 

DAU can be split into 3 exclusive user states. For example
         `DAU = Regular user v3 DAU +  New or resurrected v3 DAU + (Other v3 DAU)`, 
where 
        `other` = `(not is_regular_user_v3) and (not is_new_or_resurrected_v3)`

We can associate change in DAU with change in user state DAU; but not for WAU or MAU. As a user can change user state over a period of time, following equality won’t establish
          `MAU = Regular user v3 MAU +  New or resurrected v3 MAU + (Other v3 MAU)`, 

The left side will always be less than the right side, and the wider the window the larger the difference will be (as the higher the chance that a user is going to change user state); as seen in [here](https://sql.telemetry.mozilla.org/queries/73184/source#183289) and [here](https://sql.telemetry.mozilla.org/queries/73185/source#183288)

xluo-ds: Talked to Felix and he has clarified my above concern on the breakdown of WAU and MAU per user states.

felixlawrence:
It's probably worth linking to [this document exploring the options for WAU/MAU](https://docs.google.com/document/d/1OpejuD-wtViUX4nwddmaIHF85hU8WknjJBT-zjxxgkE/) in this PR more prominently than in a commit message.

I made a few minor changes following a discussion with @xluo-ds - if there's no other required changes then I'll merge this after she formally gives r+.

PR #531, opened 2020-07-24
Remove derived tables from choosing a dataset

acmiyaguchi:
This removes the `Main Ping Derived Datasets` section from the `Choosing a Desktop Datasets` section. Most of the markdown references were broken, and contains out-of-date material. The analysis section of the `ping_into.md` document provides enough direction to derived datasets after introducing the ping tables which should be the go-to source.


acmiyaguchi:
I'm not entirely sure how useful the information is from this page. It only includes a subset of the derived datasets in the side-bar.

wlach:
> I'm not entirely sure how useful the information is from this page. It only includes a subset of the derived datasets in the side-bar.

This page is intended for someone who is very new to telemetry and isn't familiar with the distinction between a ping table and a derived dataset. I think providing a small number suggestions for commonly used derived datasets here (along with some explanatory text) is better than nothing.

src/concepts/choosing_a_dataset.md
```diff
@@ -12,39 +12,40 @@ If you want to know whether a causal link occurs between two events, you can lea
...
+### Clients Last Seen
 
-## `main_summary`
+{{#include ../datasets/bigquery/clients_last_seen/intro.md}}
 
-{{#include ../datasets/batch_view/main_summary/intro.md}}
+See the [`clients_last_seen` reference]( ../datasets/bigquery/clients_last_seen/reference.md) for more information.
 
-## `first_shutdown_summary`
+### Main Summary
```
wlach, line 37: I don't think we should introduce main summary to new users, as we're actively discouraging people from using it.

src/concepts/choosing_a_dataset.md
```diff
@@ -12,39 +12,40 @@ If you want to know whether a causal link occurs between two events, you can lea
...
+{{#include ../datasets/batch_view/main_summary/intro.md}}
 
-## `client_count_daily`
+See the [`main_summary` reference]( ../datasets/batch_view/main_summary/reference.md) for more information.
 
-{{#include ../datasets/obsolete/client_count_daily/intro.md}}
+## Other Datasets
 
-# Other Datasets
+### Socorro Crash Reports
```
wlach, line 45: Let's skip this too -- it's a crash dataset that only a few specialists would care about.

PR #534, opened 2020-07-27
Fix link to getting help

wlach:


akkomar:
Spell- and link checker are not happy :(

wlach:
@akkomar annoyingly, vscode seems to be messing with my markdown in a way the spellchecker doesn't like. should be fixed now, have another look?

PR #536, opened 2020-07-29
Bug 1364170 - Fill out event dataset documentation

wlach:
* Add introduction with some salient facts about why to use it over the
  raw events table
* Add a simple example


badboy:
Do these docs apply the same for Glean tables?

wlach:
> Do these docs apply the same for Glean tables?

No, this a derived dataset based on the Firefox event ping. I have to confess I don't know a ton about Glean events yet (aside from the fact that they're usually contained in an events ping).

src/datasets/batch_view/events/reference.md
```diff
@@ -1,13 +1,34 @@
 # Events
 
-Events data submitted as part of the main ping.
+This derived dataset makes it easier to analyze the [event ping][event_ping].
```
wlach, line 3: @badboy's comment made me realize we should call this out explicitly.

```suggestion
This derived dataset makes it easier to analyze the Firefox [event ping][event_ping].
```

src/datasets/batch_view/events/reference.md
```diff
@@ -1,13 +1,34 @@
...
+It has the following advantages over accessing the raw ping table (`telemetry.event`):
 
-This documentation is a work in progress. Work is
-being tracked in [bug 1364170](https://bugzilla.mozilla.org/show_bug.cgi?id=1364170)
+- There is no need to `UNNEST` the `events` column: this is already done for you.
+- You don't have to introspect a BigQuery structure to find out which process emitted an event: you can query the `event_process` column directly.
```
tdsmith, line 7: This might state the advantage more clearly:

```suggestion
- You don't have to know which process type emitted your event. If you care, you can query the `event_process` column.
```

wlach:
Sounds good! I incorporated your suggestion.

PR #537, opened 2020-07-29
Update query scheduling information

scholtzan:
Updates links and the query scheduling section

### DIFF
--- a/src/cookbooks/bigquery-airflow.md
+++ b/src/cookbooks/bigquery-airflow.md
@@ -54,11 +54,11 @@ bqetl_clients_ids:   # name of the DAG; must start with bqetl_
     retry_delay: 30m
 ```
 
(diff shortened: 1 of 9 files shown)


src/cookbooks/bigquery-airflow.md
```diff
@@ -27,46 +26,44 @@ WHERE
...
-```python
-from airflow import models
-from utils.gcp import bigquery_etl_query
-
-default_args = {
-    ...
-}
-
-dag_name = 'client_ids'
+To schedule the create query, create a `/sql/example/client_ids/metadata.yaml` file with the following content:
```
scholtzan, line 29: I guess this will be replaced soon with the doc describing how to create new datasets. But in the meantime it's still nice to have the docs relatively up-to-date

PR #541, opened 2020-08-07
Introduce Jetstream tables

tdsmith:
Describes the tables produced by Jetstream.

Closes https://github.com/mozilla/jetstream/issues/136.


src/datasets/jetstream.md
```diff
@@ -0,0 +1,243 @@
...
+`statistics_<slug>_{day, week, overall}_<index>`
+
+A view is also created that concatenates all statistics tables for an
+experiment of a given period type, named like:
+
+`statistics_<slug>_{daily, weekly, overall}`
+
+Statistics tables have the schema:
+
+<table>
```
wlach, line 54: Can we use a markdown table here (and elsewhere in the text), instead of HTML? I think that'd be much easier to review and maintain.

src/datasets/jetstream.md
```diff
@@ -0,0 +1,243 @@
...
+</tr>
+<tr class="even">
+<td>`comparison_to_branch`</td>
+<td>STRING</td>
+<td>If this row represents a comparison between two branches, this row describes which branch is being compared to. For simple A/B tests, this will be “control.”</td>
+</tr>
+<tr class="odd">
+<td>`ci_width`</td>
+<td>FLOAT64</td>
+<td>A value between 0 and 1 describing the width of the confidence interval represented by the lower and upper columns. Valued at 0.95 for 95% CIs.</td>
```
wlach, line 96: I think we can just use "confidence interval", rather than the abbreviation. Although... maybe I'm just revealing my ignorance but I find the use of "width" here confusing. Is this referring to a confidence level?

```suggestion
<td>A value between 0 and 1 describing the width of the confidence interval represented by the lower and upper columns. For example, 0.95 represents a confidence interval of 95%.</td>
```

src/datasets/jetstream.md
```diff
@@ -0,0 +1,243 @@
...
+</tr>
+</tbody>
+</table>
+
+Each combination of `(segment, metric, statistic, parameter, comparison,
+comparison_to_branch, ci_width)` uniquely describes a single data
+point.
+
+To extract the mean of `active_hours` for each branch from a weekly
+statistics view v, you could run the query:
```
wlach, line 126: This is pretty abstract. Can you make this more concrete and relatable by using an example from a real table?

src/datasets/jetstream.md
```diff
@@ -0,0 +1,243 @@
...
+```
+
+This query would return a row for each week of the experiment containing
+an estimate of the absolute difference between the treatment and control
+branches for the segment containing all users.
+
+The available segments in a table should be derived from inspection of
+the table.
+
+Jetstream’s Github wiki has a description of each statistic and
```
wlach, line 174: Link?

src/datasets/jetstream.md
```diff
@@ -0,0 +1,243 @@
+Statistical summaries of telemetry data from experiments run in Mozilla
```
wlach, line 1:
```suggestion
# Jetstream Datasets

Statistical summaries of telemetry data from experiments run in Mozilla
```

src/datasets/jetstream.md
```diff
@@ -0,0 +1,243 @@
...
+products are provided by Jetstream. These summaries are published to
+BigQuery and serve both as the substrate for the result visualization
+platform and as a resource for data scientists.
+
+Jetstream runs as part of the nightly ETL job. The tables are published
```
wlach, line 6: Generally for datasets we put the scheduling information at the bottom in a seperate section called "scheduling":

https://docs.telemetry.mozilla.org/datasets/batch_view/clients_daily/reference.html#scheduling

Can you use also provide a link to whatever code (in telemetry-airflow) which is doing the scheduling?

.spelling
```diff
@@ -115,6 +116,7 @@ filesystem
 firefox-data-docs
 FirefoxReality
 FireTV
+FLOAT64
```
wlach, line 119: Any reason you didn't use backticks for the types? Seems like an odd thing to add to the dictionary.

src/datasets/jetstream.md
tdsmith: This is what pandoc gave me when I asked for commonmark, which, sure enough, [does not spec tables](https://spec.commonmark.org/0.29/). ;_; I guess [pipe_tables](https://pandoc.org/MANUAL.html#tables) is the extension I need?

I should add this to the contributing guide.

.spelling
tdsmith: "int64 was already in the dictionary," which I agree is not the strongest argument

src/datasets/jetstream.md
tdsmith: I added the scheduling system with the tech specs but I want to highlight this to data scientists early.

src/datasets/jetstream.md
```diff
@@ -0,0 +1,245 @@
+# Jetstream Datasets
+
+Statistical summaries of telemetry data from experiments run in Mozilla
+products are provided by Jetstream. These summaries are published to
```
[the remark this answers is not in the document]
tdsmith: Done!

.spelling
tdsmith: Backticked the types.

src/datasets/jetstream.md
tdsmith: I don't want to promise that any specific table exists but I added a more realistic looking slug.

src/datasets/jetstream.md
tdsmith: Confidence level is correct, yes. (I try to avoid "confidence level" to avoid suggesting that the reader should feel subjectively confident.)

tdsmith:
> with clear delineation between explanatory text, examples, and references to the code that needs to be added (which need to be added)

I separated out the examples.

I am not sure what "references to the code that needs to be added (which needs to be added)" means.

src/datasets/jetstream.md
wlach: Got it. I'd still prefer to avoid the abbreviation, but I'll leave the decision to you.

tdsmith: Oh, I'd made the change :)

.spelling
```diff
@@ -60,6 +60,7 @@ checkpointed
 checkpointing
 chromehangs
 CircleCI
+CIs
```
wlach, line 63: Can take this out now

```suggestion
```

tdsmith:
Thanks for the close look!

PR #549, opened 2020-09-01
Re-format all markdown with prettier and make it an optional check

wlach:
For now we won't make it required, to reduce friction.


src/cookbooks/public_data.md
```diff
@@ -32,54 +32,56 @@ For example:
...
-                "incremental": true,
-                "incremental_export": false,
-                "review_link": "https://bugzilla.mozilla.org/show_bug.cgi?id=1414839",
-                "files_uri": "https://public-data.telemetry.mozilla.org/api/v1/tables/telemetry_derived/ssl_ratios/v1/files",
-                "last_updated": "https://public-data.telemetry.mozilla.org/api/v1/tables/telemetry_derived/ssl_ratios/v1/last_updated"
-            }
-        },
-        // [...]
+  "telemetry_derived": {
+    // dataset name
```
mreid-moz, line 36: This change seems to make the "comments" less readable.

mreid-moz, line 36:
```suggestion
    // ^ dataset name
```

src/cookbooks/public_data.md
```diff
@@ -32,54 +32,56 @@ For example:
...
+    "deviations": {
+      // table name
```
mreid-moz, line 38:
```suggestion
      // ^ table name
```

src/cookbooks/public_data.md
```diff
@@ -32,54 +32,56 @@ For example:
...
+      "v1": {
+        // table version
```
mreid-moz, line 40:
```suggestion
        // ^ table version
```

src/cookbooks/publishing_datasets.md
```diff
@@ -18,30 +18,31 @@ owners:
   - example@mozilla.com
 labels:
   application: firefox
-  incremental: true     # incremental queries add data to existing tables
-  schedule: daily       # scheduled in Airflow to run daily
+  incremental: true # incremental queries add data to existing tables
```
mreid-moz, line 21: Ditto - the comments were aligned vertically before, making it easier to read them.

mreid-moz (approved):
This looks good to me - ~there are a few places where (imo) the changes make the markdown slightly harder to read, such as whitespace around comments in code blocks and to a lesser extent table formatting which sometimes results in very long lines~. After writing this I read [`prettier`s philosophy page](https://prettier.io/docs/en/why-prettier.html) and agree that these minor issues are worth it for consistent style.

This is a nice improvement! Thanks!

src/cookbooks/publishing_datasets.md
wlach: Looks like someone else had the same opinion: prettier/prettier#8690

I don't think this is really worth worrying about. It's still quite readable IMO.

PR #550, opened 2020-09-10
Improve ping level documentation

wlach:
* Move duplicated ping_intro documentation into a single document,
  cleaning up the table of contents in the process
* Clarify distinction between Firefox pings and Glean pings
* Various other minor improvements to wording and flow


wlach:
Current state of "choosing a desktop dataset" which partially motivated this change:

image

src/concepts/choosing_a_dataset.md
```diff
@@ -8,19 +8,23 @@ This document will help you find the best data source for a given analysis. It f
...
+## Raw Ping Datasets
 
-## Raw Pings
+We receive data from Firefox users via **pings**: small JSON payloads sent by clients at specified intervals.
+There are many types of pings, each containing different measurements and sent for different purposes.
 
-{{#include ../datasets/ping_intro.md}}
+These pings are then [aggregated into ping-level datasets](../cookbooks/bigquery/querying.md#structure-of-ping-tables-in-bigquery) that can be retrieved using BigQuery.
+Pings can be a bit difficult to work with in their raw form and there are many pitfalls that are easy to fall into when using them naively.
+Where possible, you should use a derived dataset to answer your question.
```
tdsmith, line 20: Do we still believe this? `main_summary` is described as deprecated; _that_ page tells you to use `telemetry.main`. If we don't have a practice of adding new columns to `clients_daily` and `clients_last_seen` then there are classes of analysis they won't support.

I also think the "raw ping" phraseology is a throwback to a time where the storage format and access semantics were meaningfully different from the derived datasets; maybe this is out of scope for this change but it probably makes sense to just call them "ping tables" now.

src/datasets/pings.md
```diff
@@ -1,19 +1,133 @@
 # Raw Ping Data
 
+> **⚠** This article discusses pings sent by Firefox's legacy telemetry system.
```
tdsmith, line 3:
```suggestion
> **⚠** This article discusses pings sent by Firefox's legacy v4 telemetry system.
```

Firefox has had a few legacy telemetry systems; this is the one we still use. :D

src/concepts/choosing_a_dataset.md
```diff
@@ -8,19 +8,23 @@ This document will help you find the best data source for a given analysis. It f
...
 
-## Main Ping Derived Datasets
+For more information on pings and how to use them, see [Raw Ping Data](../datasets/pings.md).
 
-The [main ping] includes most of the measurements that track the performance and health of Firefox in the wild. This ping includes histograms, scalars, and events.
+## Derived Datasets
 
-In its raw form, the main ping can be a bit difficult to work with. To make analyzing data easier, some datasets have been provided that simplify and aggregate information provided by the main ping.
+Derived datasets are built using the raw ping data above with various transformations to make them easier to work with and help you avoid common pitfalls.
```
tdsmith, line 26: I guess I wonder which pitfalls derived datasets generally help you avoid. clients_daily in particular lends itself to _different_ pitfalls (once-per-session metrics are often misaggregated).

src/datasets/pings.md
```diff
@@ -31,3 +145,52 @@ As of September 28, 2018, members of the `meta` key on main pings include:
...
+
+## Analysis
+
+The large majority of analyses can be completed using only the
+[main ping][main_ping].
+This ping includes histograms, scalars, and other performance and diagnostic data.
+
+Few analyses actually rely directly on any raw ping data.
+Instead, we provide [derived datasets](./derived.md) which are processed versions of these data,
+made to be:
```
tdsmith, line 157: Definitely untrue as written :)

```suggestion
[Derived datasets](./derived.md) are processed from ping tables. They are intended to be:
```

src/datasets/pings.md
```diff
@@ -31,3 +145,52 @@ As of September 28, 2018, members of the `meta` key on main pings include:
 <!-- table generated via `scripts/new_ping_metadata_table.py > src/cookbooks/new_ping_metadata_table.md` -->
 
 {{#include ../cookbooks/new_ping_metadata_table.md}}
+
+## Analysis
+
+The large majority of analyses can be completed using only the
+[main ping][main_ping].
+This ping includes histograms, scalars, and other performance and diagnostic data.
```
tdsmith, line 153:
```suggestion
The [main ping][main_ping] includes histograms, scalars, and other performance and diagnostic data.
Since Firefox 62, it [no longer contains event data](https://bugzilla.mozilla.org/show_bug.cgi?id=1460595); events are now sent in a separate `event` ping.
```

src/datasets/pings.md
```diff
@@ -31,3 +145,52 @@ As of September 28, 2018, members of the `meta` key on main pings include:
...
+
+Few analyses actually rely directly on any raw ping data.
+Instead, we provide [derived datasets](./derived.md) which are processed versions of these data,
+made to be:
+
+- Easier and faster to query
+- Organized to make the data easier to analyze
+- Cleaned of erroneous or misleading data
```
tdsmith, line 161: Not to keep hammering this but I think we should avoid implying that e.g. clients_daily is generally "safe". I buy 1 and 2, though.

src/datasets/pings.md
```diff
@@ -31,3 +145,52 @@ As of September 28, 2018, members of the `meta` key on main pings include:
...
+
+Ping data lives in BigQuery and is accessible in [STMO][stmo];
+see the [BigQuery cookbook section](../cookbooks/bigquery.md) for more information.
+Before analyzing raw ping data,
+**check to make sure there isn't already a derived dataset** made for your purpose.
```
tdsmith, line 166:
```suggestion
you can check if a derived dataset can answer your question.
```

tdsmith (approved):
Sorry for the delay here. +1 on reducing the amount of transclusion. The organization of the content makes sense to me.

I don't at all agree that derived datasets (except for `telemetry.events`, heh) should be the first port of call for data scientists. The ping tables are both complete and more predictable. We should at least get the story straight with the main_summary doc.

src/concepts/choosing_a_dataset.md
wlach: I'd probably describe `main_summary` less as a derived dataset and more as a precursor to the ping tables. Other than that, I agree with you that the wording could be better, let's try to do that while we're here.

src/concepts/choosing_a_dataset.md
wlach: I'm not sure what the original intent was, but I was thinking mainly of pseudo-replication. How about this?

```suggestion
Derived datasets are built using the raw ping data above with various transformations to make them easier to work with and help you avoid the pitfall of [pseudo-replication](https://docs.telemetry.mozilla.org/concepts/analysis_gotchas.html#pseudo-replication).
```

src/datasets/pings.md
wlach: Maybe we should just take out 3?

```suggestion
```

src/concepts/choosing_a_dataset.md
wlach, line 20: How about this?

```suggestion
These pings are then [aggregated into ping-level datasets](../cookbooks/bigquery/querying.md#structure-of-ping-tables-in-bigquery) that can be retrieved using BigQuery.
Pings can be difficult to work with and expensive to query: where possible, you should use a derived dataset to answer your question.
```

wlach:
> I don't at all agree that derived datasets (except for `telemetry.events`, heh) should be the first port of call for data scientists. The ping tables are both complete and more predictable. We should at least get the story straight with the main_summary doc.

I'm all for providing more context and nuance here about the tradeoffs, as long as it doesn't get too confusing.

Let's handle main_summary in a separate issue/PR. It sounds like we need to provide more context about why it used to exist, and why it isn't really a derived dataset.

src/concepts/choosing_a_dataset.md
tdsmith: sgtm!

src/concepts/choosing_a_dataset.md
tdsmith: sgtm!

src/datasets/pings.md
tdsmith: sgtm!

tdsmith (approved):

PR #564, opened 2020-10-06
Fix spelling and broken bigquery-etl links

scholtzan:
The spellchecking and and link check is not run on PRs?


jklukas:
We allow failures on the link check due to frequent intermittent failures: https://github.com/mozilla/firefox-data-docs/blob/master/.travis.yml#L33-L35

Spelling should be enforced, though.

jklukas (approved):

PR #574, opened 2020-10-20
Update for Fenix release

tdsmith:
This needs more work, probably, but this is the ~minimal edit to avoid pointing people at the Nightly table for now.

aside: I love editing pipe tables by hand; it's great

### DIFF
--- a/src/concepts/choosing_a_dataset_mobile.md
+++ b/src/concepts/choosing_a_dataset_mobile.md
@@ -2,19 +2,23 @@
 
 ## Products Overview
 
-Before you perform an analysis, it is important to identify the products that you want to include. You can select any of the following mobile products:
+Mobile products may send either legacy telemetry (see "Legacy ping tables" below), Glean telemetry, or both.
 
-| Product Name            | App Name              | OS      | Notes                                  |
-| ----------------------- | --------------------- | ------- | -------------------------------------- |
-| Firefox for Android     | `org_mozilla_firefox` | Android | formerly Fenix; uses Glean (see below) |
-| Firefox Android (old)   | `Fennec`              | Android | End-of-life; replaced by above         |
-| Firefox iOS             | `Fennec`              | iOS     |                                        |
-| Focus Android           | `Focus`               | Android | Privacy browser                        |
-| Focus iOS               | `Focus`               | iOS     | Privacy browser                        |
-| Klar                    | `Klar`                | Android | German Focus release                   |
-| Firefox for Fire TV     | `FirefoxForFireTV`    | Android |                                        |
-| Firefox for Echo Show   | `FirefoxConnect`      | Android |                                        |
-| Firefox Lite            | `Zerda`               | Android | Formerly Rocket (See below)            |
+| Marketing name        | OS      | Legacy `app_name`  | Glean dataset                                                                                             | Notes                                  |
+| --------------------- | ------- | ------------------ | --------------------------------------------------------------------------------------------------------- | -------------------------------------- |
+| Firefox for Android   | Android |                    | `org_mozilla_firefox` (release), `org_mozilla_firefox_beta` (beta), `org_mozilla_fenix` (nightly)         | formerly Fenix; uses Glean (see below) |
+| Firefox Android (old) | iOS     | `Fennec`           |                                                                                                           | End-of-life; replaced by above         |
+| Firefox iOS           | iOS     | `Fennec`           | `org_mozilla_ios_firefox` (release), `org_mozilla_ios_firefoxbeta` (beta), `org_mozilla_ios_fennec` (dev) |                                        |
+| Focus Android         | Android | `Focus`            |                                                                                                           | Privacy browser                        |
+| Focus iOS             | iOS     | `Focus`            |                                                                                                           | Privacy browser                        |
+| Klar                  | Android | `Klar`             |                                                                                                           | German Focus release                   |
+| Firefox for Fire TV   | Android | `FirefoxForFireTV` | `org_mozilla_tv_firefox `                                                                                 |                                        |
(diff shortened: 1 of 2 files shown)


src/concepts/choosing_a_dataset_mobile.md
```diff
@@ -2,19 +2,19 @@
...
-| Firefox for Echo Show   | `FirefoxConnect`    | Android |                             |
-| Firefox Lite            | `Zerda`             | Android | Formerly Rocket (See below) |
-| Fenix (Firefox Preview) | `org_mozilla_fenix` | Android | Uses Glean (see below)      |
+Before you perform an analysis, it is important to identify the products that you want to include. You can select any of the following mobile products:
+
+| Product Name            | App Name              | OS      | Notes                                  |
+| ----------------------- | --------------------- | ------- | -------------------------------------- |
+| Firefox for Android     | `org_mozilla_firefox` | Android | formerly Fenix; uses Glean (see below) |
+| Firefox Android (old)   | `Fennec`              | Android | End-of-life; replaced by above         |
+| Firefox iOS             | `Fennec`              | iOS     |                                        |
```
wlach, line 11:
```suggestion
| Firefox iOS             | `Fennec`              | iOS     |  Uses Glean (see below)                              |
```

src/concepts/choosing_a_dataset_mobile.md
```diff
@@ -2,19 +2,19 @@
...
+| Focus Android           | `Focus`               | Android | Privacy browser                        |
+| Focus iOS               | `Focus`               | iOS     | Privacy browser                        |
+| Klar                    | `Klar`                | Android | German Focus release                   |
+| Firefox for Fire TV     | `FirefoxForFireTV`    | Android |                                        |
+| Firefox for Echo Show   | `FirefoxConnect`      | Android |                                        |
+| Firefox Lite            | `Zerda`               | Android | Formerly Rocket (See below)            |
```
wlach, line 17: Do any of the above use Glean?

wlach (approved):
> This needs more work, probably, but this is the ~minimal edit to avoid pointing people at the Nightly table for now.

This is good, thank you, have one suggestion re: iOS, which now uses Glean. 

 > aside: I love editing pipe tables by hand; it's great

This is no longer necessary! You can just run prettier (or better yet, get your editor to run it via format-on-save) https://github.com/mozilla/data-docs#markdown-formatting

Note that we re-run prettier on a weekly basis so no need to worry (too much) about making sure that any individual PR is formatted.

src/concepts/choosing_a_dataset_mobile.md
travis79: ### Uses Glean:

Firefox for Fire TV
Firefox for Echo Show
Lockwise Android (going away?)
Lockwise iOS (also going away?)
Firefox Reality for Android

### Planning on integrating with Glean very soon:

Focus Android
Focus iOS
Klar? (discussion of whether or not to instrument Klar is in the works)

jklukas:
We now have a canonical UDF for "prettifying" app names, including a table of transformations in the docs. See https://mozilla.github.io/bigquery-etl/mozfun/norm/#product_info-udf

I think that table should probably be the source of truth documentation, since it's used to generate the UDF logic and is guaranteed to stay up to date. We should link to that table from these docs, and make sure our terminology is consistent between the two.

That said, the `product_info` table is currently mostly focused on making sure we have all the bases covered for mobile telemetry that feeds into KPIs. Currently, Fenix is the only app for which baseline pings are the source of truth we use for KPIs. We should definitely transition KPI reporting for Firefox iOS to baseline pings, but that has not happened yet.

So, I think it would be good for the docs here to list Glean apps for the time being, but defer to https://mozilla.github.io/bigquery-etl/mozfun/norm/#product_info-udf for naming. Then we should set a goal for January to fully codify in UDF form the mapping for Glean apps, and move to Glean as source of truth for KPIs wherever possible.

src/concepts/choosing_a_dataset_mobile.md
jklukas: For completeness, there is also Reference Browser for Android that sends Glean telemetry.

Fire TV, Echo Show, Firefox Reality, and Firefox for iOS all send both `core` pings and Glean telemetry, and KPI reporting still relies on the `core` pings.

tdsmith:
@jklukas do you know how to assign these datasets to release channels?

```
  org_mozilla_ios_fennec                 
  org_mozilla_ios_firefox                
  org_mozilla_ios_firefoxbeta
```

tdsmith:
^ please disregard; Frank reminded me that the probe-info service can shed some light on these https://probeinfo.telemetry.mozilla.org/glean/repositories

src/concepts/choosing_a_dataset_mobile.md
```diff
@@ -24,15 +28,15 @@ Klar is the known release name for Focus in Germany.
...
+When consulting legacy telemetry, filter on app name to make sure that you are looking at only the app for which you want to analyze data.
 
 [fxlite]: https://github.com/mozilla-tw/FirefoxLite/blob/master/docs/telemetry.md
 [ios]: https://github.com/mozilla-mobile/telemetry-ios
 
-## Raw Pings
+## Legacy ping tables
 
-Mobile data is structured differently when compared to desktop data. Instead of sending a `main` ping, mobile has provides the following key types of pings:
+Legacy (pre-Glean) mobile data is structured differently than desktop data. Instead of sending a `main` ping, mobile has provides the following key types of pings:
```
jklukas, line 39: I think the info here on legacy `event` pings is incorrect. It's my understanding that events for mobile apps are sent in `mobile-event` and `focus-event` pings. The `event` ping is Desktop-specific.

jklukas:
I did not realize that https://probeinfo.telemetry.mozilla.org/glean/repositories contained logical channel names! This is great! We should absolutely be using this data as source of truth for building some downstream views. I'm going to write up an issue in bigquery-etl.

PR #576, opened 2020-10-21
Redirect old links to search datasets

sunahsuh:
Uses the fairly-recently-added mdbook redirect feature for old links to the search datasets


wlach (approved):
Checked with the deploy preview and this works great. 💯

PR #578, opened 2020-11-03
Update prototype project request bug template

whd:


mreid-moz (approved):
Looks good. For reference, this updates the instructions to needinfo "a member of the Data SRE team" rather than `metadave` directly.

PR #579, opened 2020-11-04
Draft documentation for DS workflow for adding to clients_last_seen

jklukas:
I will be shopping around this workflow to some Data Scientists for review. It is intended to empower them to explore new feature usage definitions in a manner that has a clear path to being included in clients_last_seen if it proves useful.

I'm coming to the conclusion that [a general `clients_all_time` or `feature_usage_all_time` table](https://docs.google.com/document/d/1pWDgOB3rvfvuwB71EKqvSJJCHpX8iGCU9bS9yVGV31w/edit#) is not feasible in the short term, but the technique is still useful for rapid prototyping of bit patterns that can be used to investigate new user segmentation, feature usage definitions, etc.

Relevant to the [Data Warehouse daily aggregations sub-project](https://docs.google.com/document/d/1Lml-hWiqhvUazjn_-sDF6TUvwAEA3jMG3LJbyIC7fEc/edit#).

### DIFF
--- a/src/cookbooks/clients_last_seen_bits.md
+++ b/src/cookbooks/clients_last_seen_bits.md
@@ -621,6 +621,89 @@ GROUP BY
   metric_date
 ```
 
+## Proposing a new bit pattern field
+
+The operational logic required to produce a `clients_last_seen` table makes
+it unwieldy for backfilling, so it has historically been difficult for data
+scientists to experiment with new bit pattern fields on their own.
+
+Below is a sample query for producing a small `clients_last_seen`-like view
+that presents an experimental usage definition. In this approach, the temporary
+analysis table we create actually stores a client's whole usage history as a
+BYTES field, and then we rely on view logic to present this as per-day windows.
+
+The example queries `main_v4` directly in order to be as generic as possible.
+The `daily` CTE below could be removed in the case that `clients_daily` already
+provides a daily measure that can be used as basis for your new bit pattern feature.
+
+```sql
+DECLARE start_date DATE DEFAULT '2018-11-01';
+DECLARE target_date DATE DEFAULT '2020-11-01';
+DECLARE target_sample_id INT64 DEFAULT 0;
+
+CREATE OR REPLACE TABLE
+  analysis.myuser_newfeature_raw
+AS
+WITH daily AS (
+  SELECT
+    DATE(submission_timestamp) AS submission_date,
+    client_id,
+    SUM(payload.processes.parent.scalars.browser_engagement_active_ticks)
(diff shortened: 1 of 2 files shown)


jklukas:
cc @felixlawrence @irrationalagent @SuYoungHong @godelstheory for feedback. Does this seem usable? Is there obvious value in defining this workflow?

irrationalagent:
This does look like it would be useful to me.

Many feature usage metrics are based on events. At first glance, it seems like this could be adapted to use the events table by aggregating LOGICAL_OR-like per client-day. Am I missing anything else there?

jklukas:
> It seems like this could be adapted to use the events table by aggregating LOGICAL_OR-like per client-day

Yes, that's quite feasible. Jesse prototyped some events-based usage criteria in https://github.com/mozilla/bigquery-etl/pull/1193 using that pattern. The path for getting such a feature integrated into clients_last_seen is unpaved, though, so there would be discussions needed to figure out what it's like to productionize (which I'm happy to shepherd if we get there).

src/cookbooks/clients_last_seen_bits.md
```diff
@@ -621,6 +621,119 @@ GROUP BY
...
+
+Below is a sample query for producing a small `clients_last_seen`-like view
+that presents an experimental usage definition. In this approach, the temporary
+analysis table we create actually stores a client's whole usage history as a
+BYTES field, and then we rely on view logic to present this as per-day windows.
+Much of the logic is boilerplate; the sections that would need to change for
+your specific new field are marked between `-- BEGIN` and `-- END` comments.
+
+The example queries `main_v4` directly in order to be as generic as possible.
+The `daily` CTE below could be removed in the case that `clients_daily` already
```
felixlawrence, line 638: FWIW, I suspect that >90% of the time, `clients_daily` will be enough

felixlawrence (approved):
Looks good to me. I think that none of my current or planned work truly needs this (though I could well be overlooking something), but it's a good thing to have in the toolbox, and is an comprehensible explanation of the tool.

godelstheory (approved):
This is already proving very useful for the core actives work, where we are constrained with a URI>=1 as active day definition.  It is a straightforward to create a `core actives` table to join upon, with long enough lookbacks for forecasting.

jklukas:
>  It is a straightforward to create a core actives table to join upon, with long enough lookbacks for forecasting.

Can you point me to this work? I'd be very interested to see how you've chosen to adapt this approach. Are you using the BYTES fields to look at more than 28 days of history from a single row?

PR #582, opened 2020-11-16
Use mozfun bits28 functions

fbertsch:
This makes copying+pasting the code accessible in any project.

### DIFF
--- a/src/cookbooks/clients_last_seen_bits.md
+++ b/src/cookbooks/clients_last_seen_bits.md
@@ -57,7 +57,7 @@ ORDER BY
   submission_date
 ```
 
-Under the hood, `days_since_seen` is calculated using the `bits28_days_since_seen`
+Under the hood, `days_since_seen` is calculated using the `bits28.days_since_seen`
 UDF which is explained in more detail later in this article.
 
 Note that the desktop `clients_last_seen` table also has additional bit pattern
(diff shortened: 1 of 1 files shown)


src/cookbooks/clients_last_seen_bits.md
```diff
@@ -57,7 +57,7 @@ ORDER BY
   submission_date
 ```
 
-Under the hood, `days_since_seen` is calculated using the `bits28_days_since_seen`
+Under the hood, `days_since_seen` is calculated using the `bits28.days_since_seen`
```
jklukas, line 60: I think we should fully qualify the names everywhere for consistency. Some of the SQL examples here still won't work because the lack the `mozfun.` prefix.

jklukas: I take it back. I think all the actual examples here are now valid 👍

jklukas (approved):

jklukas:
Thanks for doing this.

src/cookbooks/clients_last_seen_bits.md
fbertsch: Yeah, I wasn't sure about putting the `mozfun` prefix on instances that just were describing the functions (not examples). I ended up keeping the existing structure, which didn't have the dataset names

PR #587, opened 2020-12-18
Sketch out some gotchas around build ids.

chutten:


wlach (changes requested):
Thanks for this! This is valuable clarification since build id will often come up when doing analysis. However, I think we ought to be a bit more precise here:

1. Can we give a quick definition on what a build id is? Inline here is fine, but it might be worth adding "Build ID" to our terminology section: https://docs.telemetry.mozilla.org/concepts/terminology.html
2. For Firefox desktop, they *are* consistent across channels and versions, across all time (although note that the build id will be different for each platform) and have a clearly defined format (always 10 characters: <year><month><day><hour><minute><second> e.g. 20201218095607). It is only for other applications where we have no such guarantees.

src/concepts/analysis_gotchas.md
```diff
@@ -244,3 +244,15 @@ Summary of reasons for this decision:
...
 
 On desktop, one field to watch out for is headless mode (`environment.system.gfx.headless` in the main ping): if that field is set, you are for certain not working with a version of Firefox being operated by a real human. You can see an example of some client pings with this field set skewing the nightly numbers in [bug 1643341](https://bugzilla.mozilla.org/show_bug.cgi?id=1643341). An easy solution is to just filter out these types of clients in your analysis. You can see an example of this pattern in [this query](https://sql.telemetry.mozilla.org/queries/71781/source).
+
+## Build Ids
+
+Generally speaking, application versions are monotonically increasing multipart alphanumeric strings like "89.0a1" or "68.0.3".
+Build Ids are not this.
+A Build Id is a sequence of characters that identify the specific code and configuration built into a product.
+It helps identify which code changes were included in a build as well as what features may have been enabled at build time.
```
wlach, line 253: The second two sentences are saying pretty similar things. Tried to separate them out into a summary and a more detailed explanation.

```suggestion
Generally speaking, application versions are monotonically increasing multipart alphanumeric strings like "89.0a1" or "68.0.3".
Build Ids are not this.
A Build Id is a sequence of characters that is unique to a specific build of a product.
Since the application version may not vary across shipped versions (for example, a Firefox nightly version stays the same across its entire cycle), a build id helps identify which code changes were included in a build as well as what features may have been enabled for it.
```

src/concepts/analysis_gotchas.md
```diff
@@ -244,3 +244,15 @@ Summary of reasons for this decision:
...
+For example, in Firefox Desktop, the build id is the date and time the build was built in yyyymmddhhmmss format.
+A build id might be formatted in any way and contain the time or version control system revision of the code included in the build.
+
+Do not assume the build id format does not vary between products, between channels of the same product, or over time within the same channel of the same product.
+(The build id format for Firefox Desktop has been very stable over time thus far, but even it can be different for different platforms in some respin circumstances (if e.g. only one platform's builder failed)).
```
wlach, line 258: "Not not" tends to be a bit confusing, especially for non-english-native speakers. 

```suggestion
Do not assume build id's are consistent across the products we ship. A build id format may vary between products, between channels of the same product, or over time within the same channel of the same product.
The build id format for Firefox Desktop has been very stable over time thus far, but even it can be different for different platforms in some respin circumstances (if e.g. only one platform's builder failed).
```

wlach (approved):
Awesome! Thanks for doing this. r+wc

chutten:
Ack. Accursed spellcheck.

PR #592, opened 2021-01-11
Add GitHub actions

badboy:


badboy:
(woops, still testing, this should go against my test repo for now)

PR #593, opened 2021-01-11
Bug 1685724 - Build and deploy using GitHub Actions

badboy:
This migrates the existing Travis CI tasks using only GitHub-provided
GitHub Actions and some shell commands. It can't get more simple.

Builds run on pull-requests and pushes against the master branch.
Deploys only run against pushes to the master branch.

---

GitHub Actions does not have an equivalent for Travis' `allow_failures`, but that's fine: it's their own tasks. We can make the `Build` and `Spellcheck` task required with branch protections here on GitHub, that way failures still surface in PRs and are visible in the Actions tab, but don't block landing a PR.


.gitignore
```diff
@@ -2,3 +2,4 @@ _book
 node_modules
 .*
 book
+!.github
```
wlach, line 5: What's the rationale behind this line?

wlach (commented):
This looks great, just have one question before we land. It sounds like there are two followup tasks after that:

1. Update the repository configuration to require build and spellcheck to pass.
2. Update the reminder in #data-docs to suggest looking at GitHub status instead of Travis

.gitignore
badboy: `.*` excludes all dotfiles, `!.github` is then an exception to this rule (without this one had to `git add --force` the new workflow files, though changes will then be tracked regardless of this rule)

wlach: Oh, I see -- I somehow missed the line above. I'm not sure if blanket-ignoring dotfiles is a good idea, but this is in line with what's there right now so no reason to block merging on it. r+

wlach (approved):

mreid-moz (approved):
👍 ship it!

PR #597, opened 2021-01-22
Add note about `event_timestamp` to events dataset doc

sunahsuh:
Realized after Teon's question in Slack today that we don't mention `event_timestamp`


src/datasets/batch_view/events/reference.md
```diff
@@ -13,12 +13,14 @@ It has the following advantages over accessing the raw ping table (`telemetry.ev
...
 
-The `category`, `method`, `object`, `value`, and `extra` fields of the event
-are mapped to columns named `event_category`, `event_method`, `event_object`,
+The `timestamp`, `category`, `method`, `object`, `value`, and `extra` fields of the event
+are mapped to columns named `event_timestamp`, `event_category`, `event_method`, `event_object`,
 `event_string_value`, and `event_map_values`.
 To access the `event_map_values`, you can use the `mozfun.map.get_key` UDF,
 like `SELECT mozfun.map.get_key(event_map_values, "branch") AS branch FROM telemetry.events`.
 
+Please note that `event_timestamp` refers to the time in milliseconds when the event was recorded *relative to the main process start time*, while the `timestamp` column refers to the time the ping was ingested. `event_timestamp` is useful for determining relative order of events within a single session but cannot reliably be used to determine an absolute time that an event occurred due to client clock skew and other factors.
```
tdsmith, line 22: I think this is too nihilistic; we can tell from watching Remote Settings push events that most clocks are pretty good!

```suggestion
Please note that `event_timestamp` refers to the time in milliseconds when the event was recorded *relative to the main process start time* (`session_start_time`), while the `timestamp` column refers to the time the ping was ingested. `event_timestamp` is useful for determining relative order of events within a single session. Adding `event_timestamp` to `session_start_time` will allow you to approximate the absolute time an event occurred, subject to client clock skew and other factors.
```

tdsmith (approved):
ty!

PR #600, opened 2021-01-27
Remove caveats from activation metric

tdsmith:
cf https://jira.mozilla.com/browse/DS-696


jklukas (approved):

jklukas:
Note: that the prettier-check failure is unrelated to this content, so no worries there.

PR #601, opened 2021-01-29
Clarify job titles and functions

wlach:
The terminology feels a bit out of sync with how we actually do things around here.


mreid-moz (approved):
👍

PR #612, opened 2021-02-23
Fix "edit" links in DTMO to not reference master

wlach:


wlach:
See also: https://github.com/badboy/mdbook-open-on-gh/pull/2

badboy (approved):

PR #613, opened 2021-02-23
Bug 1689124 - Add a cookbook for scheduling queries

wlach:
This replaces an out-of-date cookbook on scheduling a query
in airflow (which has been replaced with documentation inside
the bigquery-etl repository).


src/SUMMARY.md
```diff
@@ -40,7 +40,7 @@
   - [Operational](cookbooks/operational/index.md)
     - [Creating a Prototype Data Project on Google Cloud Platform](cookbooks/gcp-projects.md)
     - [Creating Static Dashboards with Protosaur](cookbooks/operational/protosaur.md)
-    - [Scheduling BigQuery Queries in Airflow](cookbooks/bigquery-airflow.md)
+    - [Scheduling Queries](cookbooks/scheduling_queries.md)
```
jklukas, line 43: How would you feel about maintaining the existing `bigquery-airflow.md` filename for continuity? Existing links or bookmarks would then still get users to relevant query scheduling docs.

wlach: Good point, but I think we can handle this via a redirect. Update incoming.

src/cookbooks/publishing_datasets.md
```diff
@@ -1,13 +1,13 @@
 # Making Datasets Publicly Available
 
 Currently, only datasets and query results that are available in BigQuery and
-defined in the [bigquery-etl][bigquery_etl] repository can be made publicly available.
-See [Scheduling BigQuery Queries in Airflow](../cookbooks/bigquery-airflow.md)
-on how to create and schedule datasets. Before data can be published, a data review is
+defined in the [BigQuery-ETL][bigquery_etl] repository can be made publicly available.
```
jklukas, line 4: I'm of the opinion that styling should be `bigquery-etl` rather than BigQuery-ETL, as it refers to a technical artifact. Definitely not a hill I'm going to die on.

src/cookbooks/scheduling_queries.md
```diff
@@ -0,0 +1,27 @@
...
+
+Mozilla's Data Engineering maintains a repository called [bigquery-etl](https://github.com/mozilla/bigquery-etl) which can incrementally create datasets based on an SQL query using our Airflow infrastructure. This requires a little bit more work to set up than scheduling a query on STMO, but has the advantage of being more cost effective, reliable, and amenable to further exploration.
+
+Also, since the procedure for scheduling a query in this way is submitting a pull request against the BigQuery-ETL repository, this is an easy route to getting peer review from the extended Data Team.
+
+For more information on how to use this approach, see [A quick guide to creating a derived dataset with BigQuery-ETL](https://mozilla.github.io/bigquery-etl/cookbooks/creating_a_derived_dataset/).
+
+## Scheduling Queries using GCP
+
+Finally, [you can schedule queries using GCP](https://cloud.google.com/bigquery/docs/scheduling-queries). This is generally not recommended, [reach out to the Data team](../concepts/getting_help.md) if you think you need to do this.
```
jklukas, line 27: In reality, this is mostly reaching out to Data SRE. This is probably fine as it is. I don't have a concrete suggestion of what else to say.

jklukas (approved):
r+wc

src/cookbooks/publishing_datasets.md
wlach: I think I went with this spelling because my tutorial used it. But I don't think I used that capitalization there very consciously... happy to change it.

src/cookbooks/scheduling_queries.md
wlach: Yeah we can see if @jasonthomas has any thoughts. But based on our internal conversations, I don't think we should do much other than discourage people from doing this.

jasonthomas: If you want to add more details around bigquery scheduling you could include "This is for short-lived queries. Scheduled queries that are active for more than 30 days will be automatically unscheduled", so that users are aware that this is not an alternative to bigquery-etl.

wlach: Sounds great, I'll add that caveat.

PR #619, opened 2021-03-11
Revise documentation on how to access Glean Data

wlach:


src/concepts/glean/accessing_glean_data.md
```diff
@@ -19,23 +40,13 @@ The top-level groups are:
 - `metrics`: [Custom metrics](https://mozilla.github.io/glean/book/user/metrics/index.html) defined by the application and its libraries.
 - `events`: [Custom events](https://mozilla.github.io/glean/book/user/metrics/event.html) defined by the application and its libraries.
 
-### Built-in metrics
-
-Accessing columns from the `client_info`, and `ping_info` group is reasonably straightforward.
+#### Built-in metrics
 
-For example, to access the `client_id` of a ping, use the column `client_info.client_id`.
+[Core metrics sent with every ping](https://mozilla.github.io/glean/book/user/pings/index.html#glean-pings) are mapped to the `client_info` and `ping_info` sections.
```
acmiyaguchi, line 45: I'd consider these attributes, not metrics.

acmiyaguchi (approved):
Seems reasonable to me.

src/concepts/glean/accessing_glean_data.md
```diff
@@ -1,13 +1,34 @@
...
 
-## Selecting the correct table
+## Using the Glean Dictionary
+
+You can see a full index of how each Glean applications maps to structures in BigQuery using the [Glean Dictionary](https://dictionary.protosaur.dev/).
+
+Select the application you're interested in from the home page, then browse through the metrics and pings. You will see the appropriate table and column
+names to use when running a query. In general, you can use these types of column names directly. For example, to get a count of distinct clients in the [Fenix baseline ping](https://dictionary.protosaur.dev/apps/fenix/pings/baseline), run this:
+
+```sql
```
mcomella, line 12: I don't think example would have helped me identify my use case. For me, I would have followed the instructions and:
- Selected fenix
- Search for `top_sites_count` & select it

But I would not have known what to do with the info on the next page. The key thing I was missing was that this:

image


means that I can actually write a SELECT query on it, e.g.
```sql
SELECT
  metrics.counter.metrics_top_sites_count
FROM
  org_mozilla_fenix.metrics
WHERE ...
```

I think for your example, it might be clearer if we do one or some of:
- we explicitly mention what content on that page can be used in BigQuery
- improve the page itself to make it clearer how those can be used in queries (with an example?)
- (least preferred) change the link we send to `client_info` rather than the baseline ping docs – I feel the pattern matching is easier there (because what I'm missing is how to extract fields from a top-level select statement item)

mcomella (commented):
Thanks for writing this! I think I might still have been confused by the docs if I had seen this so I added a comment that might help clarify what was missing from my mental model and provided some suggestions on what can change to fix that. Let me know if you have questions!

I think it could be useful to also mention the `to_json_string` function as a debugging technique but it might be fine not to

src/concepts/glean/accessing_glean_data.md
wlach: This is really helpful feedback, thanks @mcomella! I'll give this a bit more thought, it didn't occur to me that the interface in the glean dictionary didn't make this obvious.

src/concepts/pipeline/glean_data.md
```diff
@@ -21,19 +22,8 @@ The top-level groups are:
...
--- Count unique Client IDs observed on a given day
-SELECT
-  count(distinct client_info.client_id)
-FROM
-  org_mozilla_fenix.baseline
-WHERE
-  date(submission_timestamp) = '2019-11-11'
-```
+[Core attributes sent with every ping](https://mozilla.github.io/glean/book/user/pings/index.html#glean-pings) are mapped to the `client_info` and `ping_info` sections. For example,
+the client id is mapped to a column called `client_info.client_id`.
```
acmiyaguchi, line 26: May be useful to link directly to the glean documentation for these:

https://mozilla.github.io/glean/book/user/pings/index.html#the-ping_info-section
https://mozilla.github.io/glean/book/user/pings/index.html#the-client_info-section

wlach:
Ok, I took another pass at this. On reflection, I think much of the material in the old cookbook is really reference material and has no place in a cookbook. While doing this, I actually refactored this into two documents:

1. A tutorial on accessing Glean data, using some of the suggestions @mcomella provided
2. A reference on how glean data gets mapped to the data pipeline

I also found a weird gotcha while doing this (see mozilla/glean-dictionary#469), which I documented in the tutorial. Might try to get some more feedback on this before I land.

src/cookbooks/accessing_glean_data.md
```diff
@@ -0,0 +1,36 @@
+# Accessing Glean Data in BigQuery
+
+This document describes how to access Glean data using BigQuery, such as in [Redash](https://sql.telemetry.mozilla.org).
+This is intended for in-depth analysis: many simple questions can be more easily answered with either [GUD](../tools/interfaces.md#mozilla-growth--usage-dashboard-gud) or [GLAM](../tools/interfaces.md#glean-aggregated-metrics-dashboard-glam).
```
acmiyaguchi, line 4: nit: passive voice

```suggestion
This is intended for in-depth analysis: [GUD](../tools/interfaces.md#mozilla-growth--usage-dashboard-gud) or [GLAM](../tools/interfaces.md#glean-aggregated-metrics-dashboard-glam) can answer many simple questions.
```

src/cookbooks/accessing_glean_data.md
```diff
@@ -0,0 +1,36 @@
...
+  You can select which channel you want to view information for and the table name will update accordingly.
+
+## Writing a Query
+
+With this information in hand, you can now proceed to writing a query. For example, to get the
+average of this metric on the first of January, you could write something like this:
+
+```sql
+SELECT AVG(metrics.metrics.counter.metrics_top_sites_count)
+FROM org_mozilla_fenix.metrics
```
acmiyaguchi, line 30: I'd actually recommend aliasing the table e.g. `org_mozilla_fenix.metrics t` because metrics.metrics.counter.[...]` is ugly. The following section could explain why the table name aliases the column name, and give the alternative of fully qualifying the column name.

src/cookbooks/accessing_glean_data.md
```diff
@@ -0,0 +1,36 @@
...
+
+You can use the [Glean Dictionary](https://dictionary.protosaur.dev/) to access these mappings
+when writing queries. For example, say you wanted to get a count of top sites as measured in
+Firefox for Android. In this case you would:
+
+- Go to the [Glean Dictionary](https://dictionary.protosaur.dev) home page.
+- Navigate to the [Firefox for Android application](https://dictionary.protosaur.dev/apps/fenix)
+- Under metrics, search for "top", select [`metrics.top_sites_count`](https://dictionary.protosaur.dev/apps/fenix/metrics/metrics_top_sites_count).
+- Scroll down to the bottom. Under BigQuery, you should see an entry like: "In `org_mozilla_fenix.metrics` as `metrics.counter.metrics_top_sites_count`".
+  The former corresponds to the table name whilst the latter corresponds to the column name
```
acmiyaguchi, line 20: nit: punctuation
```suggestion
  The former corresponds to the table name whilst the latter corresponds to the column name.
```

src/concepts/pipeline/glean_data.md
```diff
@@ -21,19 +22,8 @@ The top-level groups are:
 
 ### Built-in metrics
```
acmiyaguchi, line 23: Rename this section. Idea: `Ping and Client Info sections`

src/cookbooks/accessing_glean_data.md
wlach: Makes sense. I updated this section to give both options, with the table alias as the first one.

PR #623, opened 2021-03-19
Fix query in glean data cookbook

acmiyaguchi:
Small nit from #619


wlach (approved):
🤦 can't believe I missed that. Thanks!

PR #624, opened 2021-03-22
Move "Accessing Additional Properties" under BigQuery

wlach:
It's really a BigQuery-specific topic, so let's put it there.


chutten (approved):
It's only useful if it can be found. Happy to have this be wherever it belongs.

PR #626, opened 2021-03-30
Bug 1701712 Add docs about events_1pct

jklukas:


src/datasets/batch_view/events/reference.md
```diff
@@ -1,6 +1,7 @@
 # Events
 
-This derived dataset makes it easier to analyze the Firefox [event ping][event_ping].
+The `telemetry.events` and `telemetry.events_1pct `derived datasets
```
wlach, line 3:
```suggestion
The `telemetry.events` and `telemetry.events_1pct` derived datasets
```

jklukas:
https://github.com/mozilla/data-docs/pull/627 will fix the link-check task.

wlach (approved):
Thanks! r+wc

I feel like it would be worth noting the caveat about how clustering won't reduce the size estimate somewhere, possibly in: https://docs.telemetry.mozilla.org/cookbooks/bigquery/optimization.html

However, that can definitely happen in a followup. Maybe we can make a note of it in our upcoming doc day.

PR #634, opened 2021-04-29
Use a consistent format (STMO#<number>) for all links to STMO

wlach:
This keeps the naming consistent and also provides a hint about how
old / new a query might be.


wlach:
I am suspecting that the right way to solve this is actually an mdbook preprocessor which can autoconvert STMO (or other) links to a nicer format for us. Converting this to a draft so I can work on that separately (have something almost working but it's past my end of day)

badboy:
I picked this up again & re-applied that across all pages, and also added a note to the style guide.
I did not automate the process, I do think this might still be worth to land now though. A more automatic way could be adopted later.

PR #641, opened 2021-05-04
Update example query for BQ CLI

rebecca-burwei:
I'm onboarding to the Data Team, just tried out these queries and saw that one of the field names was no longer valid.


jklukas (approved):
Thanks, @rebecca-burwei

And welcome!

PR #646, opened 2021-06-11
Some improvements to the introductory material

wlach:
Some of the material was written with previous content organization
in mind and is rather out of date. This is an incremental step towards
rewriting the introductary section to be more helpful and remove
deprecated content.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,7 @@
 
 [![Build Status](https://github.com/mozilla/data-docs/workflows/Build/badge.svg)](https://github.com/mozilla/data-docs/actions?query=workflow%3ABuild)
 
-This document is intended to help Mozilla's developers and data scientists
-analyze and interpret the data gathered by the Firefox Telemetry system.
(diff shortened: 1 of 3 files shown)


rebecca-burwei (approved):
Looks good to me :) One suggestion for the future: might be nice to have a tutorial / example showing how to go from the probe dictionary to finding the data for that probe in a dataset. In particular, once I have a probe I'm interested in:

- Which dataset(s) does that probe's data appear in?
- How do I select that data from the dataset? (e.g., does the probe name become a field name, a key name or something else in the dataset?)
- For keyed scalars, how do I find what the keys mean?
- How to sort out issues with time (is this probe still active?, time periods when the probe was broken, etc)

wlach:
Thanks @rebecca-burwei! I filed a bug for tracking this work and put your suggestion there:

https://bugzilla.mozilla.org/show_bug.cgi?id=1716649#c1

PR #647, opened 2021-06-11
Refresh experiment docs

tdsmith:
Remove or flag obsolete documentation, provide a better overview landing page, link out to Nimbus docs where appropriate


src/SUMMARY.md
```diff
@@ -82,10 +82,10 @@
     - [SSL Ratios (public)](datasets/other/ssl/reference.md)
     - [Telemetry Aggregates](datasets/batch_view/telemetry_aggregates/reference.md)
     - [GLAM Aggregates](datasets/glam.md)
-  - [Experiment Datasets](tools/experiments.md)
+  - [Experiment Datasets](concepts/experiments.md)
```
wlach, line 85: It feels like this document would actually better live in the section above ("data platform reference").

It's probably worth having a section here especially for experiments datasets, but we can make the table of contents a stub (like we do for derived datasets, above: https://docs.telemetry.mozilla.org/datasets/derived.html)

src/concepts/choosing_a_dataset.md
```diff
@@ -6,7 +6,7 @@ This document will help you find the best data source for a given analysis. It f
 - How many crashes occur each day?
 - How many users have installed a specific add-on?
 
-If you want to know whether a causal link occurs between two events, you can learn more at [tools for experimentation](../tools/experiments.md).
+If you want to know whether a causal link occurs between two events, you can learn more about [experimentation data](experiments.md).
```
wlach, line 9:
```suggestion
If you want to know whether a causal link occurs between two events, consider running an [experiment](experiments.md).
```

wlach (approved):
This is good. I have some minor organizational suggestions.

PR #663, opened 2021-09-22
Updates to list of search access points

skahmann3:
Added urlbar-searchmode and combined with alias. Added more details for urlbar, searchbar, newtab, and abouthome. Exact version numbers for newtab and abouthome deprecation still needed.

### DIFF
--- a/src/datasets/search.md
+++ b/src/datasets/search.md
@@ -17,19 +17,17 @@ complex analyses.
 
 Searches can be split into three major classes: _sap_, _follow-on_, and _organic_.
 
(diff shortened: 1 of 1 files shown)


wlach (approved):
I'm not up on any recent changes to the FIrefox source code around search, but this looks ok to me as far as the style is concerned. If you're uncertain, you could try doublechecking with the search engineering team (no sure where they hang out these days). If you've done that, feel free to land as-is (once the spelling issues are resolved).

skahmann3 (commented):
SAP updates

PR #666, opened 2021-10-04
Add old clients analysis gotcha

whd:
I'm not sure if we want to be more specific here as investigation is ongoing


jklukas (approved):
This seems like an appropriate level of detail for now.

PR #672, opened 2021-11-03
Add workgroup-specific AI notebooks access docs

whd:
https://mozilla-hub.atlassian.net/browse/DSRE-212

### DIFF
--- a/src/cookbooks/bigquery/access.md
+++ b/src/cookbooks/bigquery/access.md
@@ -139,6 +139,24 @@ To start, go to [GCP console](https://console.cloud.google.com) and make sure yo
 
 Please note that by default JupyterLab saves notebook files only locally, so they are lost if your instance is deleted. To make sure you don't lose your work, either push your files to a Git repository (via a pre-installed Git extension) or upload them to GCS (using `gsutil` command in a terminal session).
 
+#### Notebooks Access to workgroup-confidential Datasets
+
+If you are a member of a restricted access workgroup, you can provision AI notebooks in the [`mozdata GCP project`](https://console.cloud.google.com/vertex-ai/workbench/list/instances?project=mozdata&supportedpurview=project) that can read workgroup-confidential data.
+
+> **⚠** You must provision AI notebooks in `mozdata` using a nonstandard service account specific to your workgroup, see below.
+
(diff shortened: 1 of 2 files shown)


src/cookbooks/bigquery/access.md
```diff
@@ -139,6 +139,24 @@ To start, go to [GCP console](https://console.cloud.google.com) and make sure yo
...
+
+> **⚠** You must provision AI notebooks in `mozdata` using a nonstandard service account specific to your workgroup, see below.
+
+When you create a notebook server, under "Advanced Options" / "Permissions", deselect "Use Compute Engine Default Service Account" and replace it with the service account associated with your workgroup. You may need to type this service account manually as it will not be available from a drop-down menu to all users. The ID of the service account matches the following pattern:
+
+`WORKGROUP-SUBGROUP@mozdata.iam.gserviceaccount.com`
+
+For example, if you are member of `workgroup:search-terms/aggregated`, use `search-terms-aggregated@mozdata.iam.gserviceaccount.com`.
+
+This notebook server should have access to any restricted access datasets that are accessible to `workgroup:search-terms/aggregated`. Additionally, this notebooks server will not have write access to the standard `mozdata.analysis` dataset, but will instead have write access to a workgroup-specific dataset that looks like the following:
```
jklukas, line 154: Does the user creating this notebook server have control over who is granted access to the server? Do we need to document here that access to the server needs to match the access that the server itself has been granted?

whd: > Does the user creating this notebook server have control over who is granted access to the server? 

If you're referring to the technical capability of a user to modify a running server (which under the hood is simply a GCE instance) in such a way as to allow arbitrary users to access the server, yes. If you're referring to who is granted access as a matter of course (e.g. via the notebooks proxy URL): no, a user is restricted to launching notebooks servers associated with a workgroup/subgroup and anybody in that group will have access to their server (or just themselves if using single user mode). Some discussion of this in https://mozilla-hub.atlassian.net/browse/DSRE-212?focusedCommentId=470828, in particular that we're relying on "additional procedural restrictions on getting access to data in the first place" and auditing to discourage arbitrary server modifications.

> Do we need to document here that access to the server needs to match the access that the server itself has been granted?

I can add a section to clarify the above if you think it's worthwhile, but the assumption is analysts have already signed agreements much more formal than this that would discourage modifying servers in this manner.

jklukas (approved):

whd:
@rebecca-burwei can you try these instructions out and let me know if you run into any issues?

jklukas:
These instructions have been validated.

PR #674, opened 2021-11-08
Add a brief introduction to Looker

wlach:
This tutorial attempts to introduce some basic concepts around Looker
and ways to start exploring. More in-depth documentation and tutorials
forthcoming (hopefully).


src/cookbooks/looker/intro.md
```diff
@@ -0,0 +1,62 @@
...
+
+You can access Mozilla's instance of Looker at [`mozilla.cloud.looker.com`](https://mozilla.cloud.looker.com).
+
+## Getting Started
+
+### Front page
+
+By default, on the front page you will see a list of default folders, which contain links to dashboards.
+These are organized by project.
+Of particular note is the "Mozilla Metrics" folder, which includes several Data-produced and vetted dashboards like the Firefox Corporate KPI Dashboard.
```
scholtzan, line 22:
```suggestion
Of particular note is the ["Mozilla Metrics"](https://mozilla.cloud.looker.com/folders/63) folder, which includes several Data-produced and vetted dashboards like the Firefox Corporate KPI Dashboard.
```

src/cookbooks/looker/intro.md
```diff
@@ -0,0 +1,62 @@
...
+
+![](../../assets/Looker_screenshots/front-page.png)
+
```
scholtzan, line 25: I wonder if we should mention namespaces (=functional group represented within Looker) here? Since namespaces essentially result in new folders getting added? And maybe link to https://mozilla-hub.atlassian.net/jira/software/c/projects/LOOK/issues for requesting new namespaces?

scholtzan (approved):
Looks good. I wonder if we should migrate the existing Mana page for Looker to dtmo? I don't think there is any sensitive information on here: https://mana.mozilla.org/wiki/display/DATA/Developing+in+Looker

src/cookbooks/looker/intro.md
wlach: Linking is a good suggestion, but we should be probably remove the quotes in that case. Will do that in my branch

wlach:
> Looks good. I wonder if we should migrate the existing Mana page for Looker to dtmo? I don't think there is any sensitive information on here: https://mana.mozilla.org/wiki/display/DATA/Developing+in+Looker

Good idea, though I think most of that doesn't belong in an introductory document (much of it is basically reference material). Let's do that as a followup.

src/cookbooks/looker/intro.md
wlach: I think the namespaces concept is probably better introduced in a reference document. Let's do that as a followup

PR #675, opened 2021-11-16
Add 404 analysis gotcha

whd:
There are some SRE incident docs but they might be private. We didn't create an internal issue for this except the DS ticket to analyze impact so I've linked to that. We could perhaps alternatively link to https://status.cloud.google.com/incidents/6PM5mNd43NbMqjCZ5REh directly.


jklukas (approved):

mreid-moz (commented):
👍

PR #676, opened 2021-11-19
vscode recommended settings

wlach:
* Add prettier and spell-right as recommended extensions
* Make format-on-save the default

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,6 @@
+{
+    "recommendations": [
+        "ban.spellright",
+        "esbenp.prettier-vscode"
+    ]
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    "editor.formatOnSave": true
+}
+  


jklukas (commented):
I don't personally use vscode, so don't have much context here.

My initial reaction, is that these sorts of preferences are generally more appropriate for individuals devs to set as global defaults on their individual machines rather than checking in these prefs to a repository. I worry about this becoming a forum for bikeshedding or for tech debt if we add support for more tools.

wlach:
> I don't personally use vscode, so don't have much context here.
> 
> My initial reaction, is that these sorts of preferences are generally more appropriate for individuals devs to set as global defaults on their individual machines rather than checking in these prefs to a repository. I worry about this becoming a forum for bikeshedding or for tech debt if we add support for more tools.

Yeah, my motivation was to make the edit-in-github experience better (try pressing `.` in this repo) but in practice I'm not sure if that's an avenue really worth emphasizing (VSCode's git commit workflow is not exactly intuitively). Let's skip this for now. Perhaps in the future we can use the same approach we did in bigquery-etl: https://github.com/mozilla/bigquery-etl/pull/1929

PR #691, opened 2022-01-27
Updated analytics diagram with word logos

ANich:


sean-rose (changes requested):
Those changes are definitely helpful, thanks!

For Colab, how about using a logo that says "Colab" rather than just "Co"? ([example](https://miro.medium.com/max/1400/1*Lad06lrjlU9UZgSTHUoyfA.png))

sean-rose (approved):

sean-rose:
btw, with how you're generating the diagram, if PNG is an option I think that might be a better format for this than JPEG.

ANich:
> btw, with how you're generating the diagram, if PNG is an option I think that might be a better format for this than JPEG.

Seems like that's an existing [feature request](https://community.miro.com/wish-list-32/export-transparent-png-s-1865) for Miro. Currently supports JPG and a Vector PDF

PR #692, opened 2022-02-07
Expand the SSL dataset docs' example

chutten:
The example was mathematically correct but gave a false impression of how to use the dataset.
Can't forget to unnormalize the result, or you'll get weird figures.
Also, emphasize that the normalization is per-`submission_date`.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -359,6 +359,7 @@ unicode
 unmanaged
 unnest
 unnesting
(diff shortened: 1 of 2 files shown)


Dexterp37 (approved):
Thanks for fixing the docs. R+ with the spellcheck addressed!

PR #694, opened 2022-02-15
add gotcha section

xluo-ds:
Add gotcha section around potential mismatch b/t `search_aggregates` and `search_clients_engines_sources_daily` that Jeff mentioned.


jklukas (approved):
Looks good. It would also be appropriate to document this problem more generally on https://github.com/mozilla/data-docs/blob/main/src/concepts/analysis_gotchas.md

PR #697, opened 2022-03-10
Add subscription platform emails

clouserw:


jklukas:
The spell-check task is failing here, probably because `Stripe` (capitalized) needs to be added to [the](https://github.com/mozilla/data-docs/blob/main/.spelling)

clouserw:
Thanks.  I added it but it's still broken.  Is there a place to see more verbose output from the spellchecker?

badboy:
> Thanks. I added it but it's still broken. Is there a place to see more verbose output from the spellchecker?

You can run spellchecking locally using `script/spell_check.sh`, which will give you colored output:

image

Unfortunately the color is lost on GhA and there's no way to tell mdspell to disable color and use some other output.

tl;dr: add `webhook` to `.spelling` and you're all good.

clouserw:
Thanks @badboy .  It really helps to have the color!

jklukas:
Thanks for this updates. Sorry that the review cycle took so long.

PR #700, opened 2022-03-25
Create real_time_search.md

xluo-ds:
Add a section for pulling real time search metrics

### DIFF
--- a/src/cookbooks/real_time_search.md
+++ b/src/cookbooks/real_time_search.md
@@ -1,4 +1,3 @@
-
 # Search metrics
 
 Search metrics are delivered in main pings and often reflect usage over the course of many previous hours, so there is inherent delay (which makes it more reasonable to call `introday` search metrics instead of `almost realy-time` search metrics. Below is the query to give the hourly `sap` search metrics per each major search engine since the start of current day in `CA` based on `1%` of the sample from `telemetry_live` dataset. 
--- a/src/cookbooks/real_time_search.md
+++ b/src/cookbooks/real_time_search.md
@@ -1,6 +1,6 @@
 # Search metrics
 
-Search metrics are delivered in main pings and often reflect usage over the course of many previous hours, so there is inherent delay (which makes it more reasonable to call `introday` search metrics instead of `almost realy-time` search metrics. Below is the query to give the hourly `sap` search metrics per each major search engine since the start of current day in `CA` based on `1%` of the sample from `telemetry_live` dataset. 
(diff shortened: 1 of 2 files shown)


src/cookbooks/real_time_search.md
```diff
@@ -0,0 +1,24 @@
+
+# See Real-time search metrics
+
+It's easy to get almost real-time search metrics with `telemetry_live` dataset. Below is the query to give the hourly `sap` search metrics per each major search engine since the start of current day in `CA` based on `1%` of the sample.
```
jklukas, line 4: For event pings, it would be reasonable to call the live tables "almost real-time", but I think it's misleading for search metrics. Search metrics are delivered in main pings and often reflect usage over the course of many previous hours, so there is inherent delay.

I would suggest maybe calling this "intraday search metrics".

jklukas (changes requested):

src/cookbooks/real_time_search.md
xluo-ds: gotcha, updated.

xluo-ds (commented):
fix the spelling error

src/cookbooks/real_time_search.md
```diff
@@ -0,0 +1,25 @@
+
+# Search metrics
```
badboy, line 2:
```suggestion
# Search metrics
```

src/cookbooks/real_time_search.md
```diff
@@ -0,0 +1,25 @@
...
+
+Search metrics are delivered in main pings and often reflect usage over the course of many previous hours, so there is inherent delay (which makes it more reasonable to call `introday` search metrics instead of `almost realy-time` search metrics. Below is the query to give the hourly `sap` search metrics per each major search engine since the start of current day in `CA` based on `1%` of the sample from `telemetry_live` dataset.
```
badboy, line 4:
```suggestion
Search metrics are delivered in main pings and often reflect usage over the course of many previous hours, so there is inherent delay (which makes it more reasonable to call `introday` search metrics instead of `almost real-time` search metrics. Below is the query to give the hourly `sap` search metrics per each major search engine since the start of current day in `CA` based on `1%` of the sample from `telemetry_live` dataset.
```

(trailing whitespace makes `prettier` unhappy.)
I usually advise to use [semantic linebreaks](https://sembr.org/) but either is fine.

xluo-ds:
Hey, apologies on this consistent failure. I tried to follow the spell checking mentioned [here](https://github.com/mozilla/data-docs) but failed on my laptop. Talked to Anna, she may have a tutorial/walk-through to reduce the friction in updating DTMO and bigquery-etl in general.

xluo-ds:
@fbertsch I think the fix to CI with the comparison of the change made debug much easier, and I have been able to figure out and solve all the failures to pass the checks. Thank you!

fbertsch (approved):
That's great! Let's get this added!

PR #708, opened 2022-04-14
Add note about BQ access requests

whd:
https://bugzilla.mozilla.org/show_bug.cgi?id=1749085#c3
https://mozilla-hub.atlassian.net/browse/DSRE-681

### DIFF
--- a/src/cookbooks/bigquery/access.md
+++ b/src/cookbooks/bigquery/access.md
@@ -159,6 +159,8 @@ In the example above this maps to `mozdata.search_terms_aggregated_analysis`.
 
 ## BigQuery Access Request
 
+> **⚠**  If you are a member of certain staff groups such as Data Engineering or Data Science you do not need to request BigQuery access specifically; it is automatically granted.
+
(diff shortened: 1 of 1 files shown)


src/cookbooks/bigquery/access.md
```diff
@@ -159,6 +159,8 @@ In the example above this maps to `mozdata.search_terms_aggregated_analysis`.
 
 ## BigQuery Access Request
 
+> **⚠**  If you are a member of certain staff groups such as Data Engineering or Data Science you do not need to request BigQuery access specifically; it is automatically granted.
```
sean-rose, line 162: "_certain staff groups such as..._" indicates there are other groups besides Data Engineering and Data Science that are automatically granted BigQuery access, which might leave non-DE/DS people wondering if they're a member of those other groups.  I'd suggest either listing all such groups, or making the statement more definitive about the DE & DS groups, like:
```suggestion
> **⚠**  If you are a member of the Data Engineering or Data Science staff groups you do not need to request BigQuery access specifically; it is automatically granted.
```

Also, is there any way for people to empirically check whether they're a member of those staff groups?

sean-rose (approved):
r+wc

src/cookbooks/bigquery/access.md
whd: I left this vague because there are other groups besides Data Engineering and Data Science that are automatically granted BigQuery access, but keeping the list up to date here is not reasonable. There is a way to check whether you're a member of a staff group that has access to mozilla-confidential data, which is by querying mozilla-confidential data from the mozdata project. The flow that would cut down on extraneous requests the most would include telling people to check whether they can access data before filing a request.

Given there are plans to redo permissions to obviate the vast majority of requests via https://mozilla-hub.atlassian.net/browse/DSRE-544 anyway, I think the most reasonable thing to do here is to make the statement definitive as to DE and DS and remove this entire section when DSRE-544 is implemented.

PR #716, opened 2022-05-09
add "urlbar" to vocabulary

rebecca-burwei:
"urlbar" will not longer be considered misspelled by automatic tests.


xluo-ds (approved):
LGTM

PR #717, opened 2022-05-10
add suggest link

xluo-ds:
need to add link to new article in SUMMARY.md according to [Adding a new article](https://docs.telemetry.mozilla.org/contributing/index.html#adding-a-new-article), that's probably why it's not shown in DTMO yet :)


rebecca-burwei (approved):
thank you!

PR #732, opened 2022-08-17
Make CI easier to use

fbertsch:
This is a follow-up from the recent DE/DS office hours. Tested here: https://github.com/fbertsch/firefox-data-docs/pull/1

- Only run checks on changed files
- Colorize outputs, including spellcheck
- Show diff of prettifier
- Ensure checks still run outside of CI

### DIFF
--- a/scripts/link_check.sh
+++ b/scripts/link_check.sh
@@ -12,5 +12,5 @@ LINKCHECK_CMD="npx markdown-link-check --quiet --verbose --config .linkcheck.jso
 if command -v parallel > /dev/null; then
   parallel -P 8 $LINKCHECK_CMD ::: "${SEARCH_ARR[@]}"
 else
-  echo "${SEARCH_ARR[*]}" | xargs -n1 $LINKCHECK_CMD
+  echo "${SEARCH_ARR[*]}" | xargs -n1 -P 8 $LINKCHECK_CMD
 fi
--- a/scripts/link_check.sh
+++ b/scripts/link_check.sh
@@ -1,11 +1,16 @@
 #!/bin/bash
+export TERM=xterm-color
 
-if command -v parallel > /dev/null; then
-  RUNNER="parallel"
-else
-  RUNNER="xargs -n1"
+SEARCH_ARR=(${@:-})
+
+if [[ -z $SEARCH_ARR ]]; then
+  SEARCH_ARR=(`find src -name "*.md"`)
 fi
 
-find src -name "*.md" |\
-  sort |\
-  $RUNNER -P 8 npx markdown-link-check --quiet --verbose --config .linkcheck.json
+LINKCHECK_CMD="npx markdown-link-check --quiet --verbose --config .linkcheck.json"
+
+if command -v parallel > /dev/null; then
+  parallel -P 8 $LINKCHECK_CMD ::: "${SEARCH_ARR[@]}"
+else
+  echo "${SEARCH_ARR[*]}" | xargs -n1 $LINKCHECK_CMD
+fi
--- a/scripts/prettier_check.sh
+++ b/scripts/prettier_check.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+export TERM=xterm-color
+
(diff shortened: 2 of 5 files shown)


src/concepts/analysis_gotchas.md
```diff
@@ -1,5 +1,8 @@
 # Analysis Gotchas
 
+ "You cannot pass," he said. The orcs stood still, and a dead silence fell. "I am a servant of the Secret Fire, wielder of the flame of Anor. You cannot pass. The dark fire will not avail you, flame of Udûn. Go back to the Shadow! You cannot pass."
```
badboy, line 3: 🤔

badboy (approved):

fbertsch:
@badboy doesn't look like this is running. First issue is this:
```
tj-actions/changed-files@v26.1 is not allowed to be used in mozilla/data-docs. Actions in this workflow must be: within a repository that belongs to your Enterprise account, created by GitHub, or matching the following: !/mozilla/**, !mozilla/**, ./**, 10up/wpcs-action@*, aws-actions/*, docker/*, pypa/gh-action-pypi-publish@v1.4.2, slackapi/slack-github-action@*, google-github-actions/*, erlef/setup-beam@v1, yesolutions/mirror-action@*, codecov/codecov-action@*.
```

And that's on a branch on this repo, not my fork. This forked repo PR doesn't seem to run tests at all.

src/concepts/analysis_gotchas.md
fbertsch: 🧙‍♂️

scripts/prettier_check.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+export TERM=xterm-color
+
+SEARCH_PATH=${@:-'src/**/*.md'}
+
+npx prettier --write --loglevel warn $SEARCH_PATH
+
+git --no-pager diff --color --exit-code src/ README.md
```
sean-rose, line 8: This is a clever solution, but IMO having this script always modify files in place would be too much of a surprise/gotcha for anyone who might run this script locally.

Suggestions:
* Have this script accept general prettier args, and if no args are passed in default to `--check src/**/*.md`.
* Rename the script `prettier.sh` (having "check" in the name would imply always doing `--check`).
* Change the CI to call this script with `--write --loglevel warn ${{ needs.get-changed-files.outputs.changed_files }}`, followed by running the `git diff` command.

sean-rose (approved):
r+wc

scripts/prettier_check.sh
fbertsch: Agreed - would definitely not be good to just have this overwrite. I might move this code over to `prettier_fix.sh`, which already overwrites

scripts/link_check.sh
```diff
@@ -1,11 +1,16 @@
...
+fi
+
+LINKCHECK_CMD="npx markdown-link-check --quiet --verbose --config .linkcheck.json"
 
 if command -v parallel > /dev/null; then
-  RUNNER="parallel"
+  parallel -P 8 $LINKCHECK_CMD ::: "${SEARCH_ARR[@]}"
 else
-  RUNNER="xargs -n1"
+  echo "${SEARCH_ARR[*]}" | xargs -n1 $LINKCHECK_CMD
```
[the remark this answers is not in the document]
fbertsch: No good reason! I'll add it back in

scripts/prettier_fix.sh
```diff
@@ -1,3 +1,8 @@
 #!/bin/bash
+export TERM=xterm-color
 
-npx prettier --write src/**/*md README.md
+SEARCH_PATH=${@:-'src/**/*.md'}
+
+npx prettier --write --loglevel warn $SEARCH_PATH
+
+git --no-pager diff --color --exit-code src/ README.md
```
sean-rose, line 8: I'd suggest moving the `git diff` into the Prettier CI command, because if someone runs this locally they might well have uncommitted changes and the diff output would be unexpected.

PR #756, opened 2022-12-20
Add docs for working with live data

scholtzan:
Addresses https://bugzilla.mozilla.org/show_bug.cgi?id=1727079 and https://bugzilla.mozilla.org/show_bug.cgi?id=1727078

Based on https://docs.google.com/document/d/1Zn8MbwQt2ANA7LyVpTAJnINDTnoX1Q7DSgjsEngwRDw/edit#heading=h.u4ctqaaww1ex

### DIFF
--- a/src/cookbooks/live_data.md
+++ b/src/cookbooks/live_data.md
@@ -2,7 +2,7 @@
 
 Live ping tables are the final destination for the telemetry ingestion pipeline. Incoming ping data is loaded into these tables approximately every 10 minutes, though a delay of up to 30 minutes is normal. Data in these tables is set to expire after 30 days.
 
-Data from live ping tables is expected to be accesses through user-facing views. The names of all views accessing live data should include the `_live` suffix (for example `monitoring.topsites_click_rate_live`). Live tables are generally clustered and partitioned by `submission_timestamp`, which allows for writing more efficient queries that filter over short time windows. When querying live data by running the same query multiple times ensure that query caching is turned off, otherwise newly arrived data might not appear in the query results.
+Data from live ping tables is expected to be accessed through user-facing views. The names of all views accessing live data should include the `_live` suffix (for example `monitoring.topsites_click_rate_live`). Live tables are generally clustered and partitioned by `submission_timestamp`, which allows for writing more efficient queries that filter over short time windows. When querying live data by running the same query multiple times ensure that query caching is turned off, otherwise newly arrived data might not appear in the query results.
 
 Of note, user-facing views for live ping tables, like the ones for historical ping tables, will not automatically be provisioned. Instead, it is expected for live views to be more curated, with each view being tied to a specific use case.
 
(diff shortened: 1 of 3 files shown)


src/cookbooks/live_data.md
```diff
@@ -0,0 +1,48 @@
+# Working with Live Data
+
+Live ping tables are the final destination for the telemetry ingestion pipeline. Incoming ping data is loaded into these tables approximately every 10 minutes, though a delay of up to 30 minutes is normal. Data in these tables is set to expire after 30 days.
+
+Data from live ping tables is expected to be accesses through user-facing views. The names of all views accessing live data should include the `_live` suffix (for example `monitoring.topsites_click_rate_live`). Live tables are generally clustered and partitioned by `submission_timestamp`, which allows for writing more efficient queries that filter over short time windows. When querying live data by running the same query multiple times ensure that query caching is turned off, otherwise newly arrived data might not appear in the query results.
```
ANich, line 5:
```suggestion
Data from live ping tables is expected to be accessed through user-facing views. The names of all views accessing live data should include the `_live` suffix (for example `monitoring.topsites_click_rate_live`). Live tables are generally clustered and partitioned by `submission_timestamp`, which allows for writing more efficient queries that filter over short time windows. When querying live data by running the same query multiple times ensure that query caching is turned off, otherwise newly arrived data might not appear in the query results.
```

src/cookbooks/live_data.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Using Materialized Views
+
+When creating a view on live data, engineers should consider the specific use cases for the view and do some analysis of the cost associated with query patterns. In particular, cost will likely be an issue if the view is going to be used for some type of monitoring that involves frequent queries on Firefox-scale data. [BigQuery offers support for materialized views](https://cloud.google.com/bigquery/docs/materialized-views-intro). Materialized views are precomputed views that periodically cache the results of a query for increased performance and efficiency. Creating a materialized view that provides the relevant aggregates will be a significant cost savings when working with a large volume of live data.
+
+When a materialized view is appropriate, it should be created in a relevant `_derived` dataset with a version suffix (following the same conventions as we do for derived tables populated via scheduled queries). It is then made user-facing via a simple `SELECT * FROM my_materialized_view` virtual view in a relevant user-facing dataset. More complex cases may need to union together multiple materialized views at this user-facing view level.
```
ANich, line 15: General comment: not a huge fan of this type of misdirection. I think in this case it's necessary since we can't create materialized views on top of views. Would be nice to be able to have the user-facing view point at the live table and the materialized view point at the user-facing view but not be a dependency

ANich (approved):
LGTM, one typo and general comment

src/cookbooks/live_data.md
scholtzan: Having the materialized view point at the user-facing view, would mean that (if the materialized view should be publicly accessible) it would need to be created twice, in `shared-prod` and `mozdata`, which could have some negative implications for cost and isn't supported by our deployment platform at the moment.

PR #760, opened 2023-01-27
Document search with ads bug fix

rebecca-burwei:


xluo-ds (approved):
thank you for taking this :)

PR #761, opened 2023-02-02
[RS-528] Add links to Suggest and Tiles gotchas

skahmann3:
[RS-528] There are several gotchas with imported Tiles & Suggest revenue data, namely around launch dates for Desktop Tiles and mobile Suggest. Given the revenue component, these gotchas are documented in an access-protected sheet rather than in long-form in DTMO.


badboy:
The prettier check failed.
You can run the formatting locally:

```
scripts/prettier_fix.sh src/datasets/other/sponsored_tiles/sponsored_tiles.md src/datasets/other/suggest/suggest.md
```

and commit the changes

rebecca-burwei (approved):
lgtm :)

PR #767, opened 2023-02-22
Helps clarify example in sql guidelines

iturgeon:
The example for `Do not include multiple arguments on one line` has a second bad example that may confuse the reader. 

The reason is it fixes a problem the bad example has but also introduces a new problem the bad example didn't have. This change would move both issues into the first bad example and fix them both in the good example.

### DIFF
--- a/src/concepts/sql_style.md
+++ b/src/concepts/sql_style.md
@@ -240,20 +240,20 @@ Do not include multiple arguments on one line.
 ```sql
 SELECT client_id, submission_date
 FROM main_summary
-WHERE
-  submission_date > '20180101'
+WHERE submission_date > '20180101'
   AND sample_id = '42'
 LIMIT 10
 ```
 
-**Bad**
+**Good**
 
 ```sql
 SELECT
   client_id,
   submission_date
 FROM main_summary
-WHERE submission_date > '20180101'
+WHERE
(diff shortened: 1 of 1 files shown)


badboy (approved):

src/concepts/sql_style.md
```diff
@@ -240,24 +240,23 @@ Do not include multiple arguments on one line.
...
   client_id,
   submission_date
 FROM main_summary
-WHERE submission_date > '20180101'
+WHERE
+  submission_date > '20180101'
   AND sample_id = '42'
 LIMIT 10
 ```
-
```
badboy, line 260: accidental line removal? we need this to make prettier happy.

src/concepts/sql_style.md
```diff
@@ -240,20 +240,20 @@ Do not include multiple arguments on one line.
...
 ```
 
-**Bad**
+**Good**
 
 ```sql
 SELECT
   client_id,
   submission_date
 FROM main_summary
```
fbertsch, line 254: `main_summary` and the `10` from `LIMIT 10` would need to move to the next lines for this to be "Good". We already have the good example above and this would simply replicate that

fbertsch:
Closing this. Feel free to file an issue if you think the examples are confusing.

PR #772, opened 2023-03-29
add and improve definitions

bochocki:
- Improve DAU definition to contain more context and links to external resources.
- Add definitions for: KPI, URI, URL

### DIFF
--- a/src/concepts/terminology.md
+++ b/src/concepts/terminology.md
@@ -61,7 +61,7 @@ A set of data, which includes ping data, derived datasets, etc.; sometimes it is
 
 Daily Active Users - The number of unique [client ids](#client-id) that we receive a [ping](#ping) from each day.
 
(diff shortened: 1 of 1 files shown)


src/concepts/terminology.md
```diff
@@ -59,7 +59,11 @@ A set of data, which includes ping data, derived datasets, etc.; sometimes it is
 
 ## DAU
 
-Daily Active Users - The number of unique profiles active on each day.
+Daily Active Users - The number of unique [client ids](#client-id) that we receive a [ping](#ping) from each day.
+
+DAU can be defined differently across different products. For example, [Firefox Desktop DAU](https://mozilla.github.io/metric-hub/metrics/firefox_desktop/#daily_active_users) only considers client ids to be "active" on a particular day if the client visits at least one [URI](#uri) and has a non-zero number of [active hours](https://mozilla.github.io/metric-hub/metrics/firefox_desktop/#active_hours). Mobile DAU, on the other hand, considers any client to be "active" on days that we receive a ping from them.
```
irrationalagent, line 64: I think for mobile its specifically, did we receive a Glean Baseline Ping - because (last I heard) we are only using the Glean reporting for KPIs right? Or maybe thats too much detail anyway...

irrationalagent (approved):

src/concepts/terminology.md
bochocki: Good call! I'll add reference to the baseline ping for mobile as well as the main ping for desktop.

PR #785, opened 2023-06-13
Change lookup to country_lookup

fbertsch:
`lookup` is a reserved keyword in BQ, so copy+pasting the code as-is results in confusing errors.

### DIFF
--- a/src/cookbooks/looker/countries.md
+++ b/src/cookbooks/looker/countries.md
@@ -54,7 +54,7 @@ Note: we use `??` as a country-code for empty country data from data sources. Th
 "Unknown Region", etc.
 
 At this point, you should check for cases where the resulting `country_code` matches `your_table.country` but does 
-not match any values in the country_lookup table - you may have discovered a new alias, in which case please add it to the list!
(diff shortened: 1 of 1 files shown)


ANich (approved):
👍 Needs format and dictionary entry

src/cookbooks/looker/countries.md
```diff
@@ -36,25 +36,25 @@ CI:
...
-    LEFT JOIN mozdata.static.country_codes_v1 lookup ON your_table.country_field = lookup.name 
+    LEFT JOIN mozdata.static.country_codes_v1 country_lookup ON your_table.country_field = country_lookup.name 
 ```
 
 Note: we use `??` as a country-code for empty country data from data sources. This will map to "Unknown Country", 
 "Unknown Region", etc.
 
 At this point, you should check for cases where the resulting `country_code` matches `your_table.country` but does 
-not match any values in the lookup table - you may have discovered a new alias, in which case please add it to the list!
+not match any values in the country_lookup table - you may have discovered a new alias, in which case please add it to the list!
```
badboy, line 57:
```suggestion
not match any values in the `country_lookup` table - you may have discovered a new alias, in which case please add it to the list!
```

PR #794, opened 2023-08-31
Update SQL style guide about indenting join conditions

sean-rose:
mozilla/bigquery-etl#4223 is changing `bqetl format` to indent join conditions.

### DIFF
--- a/src/concepts/sql_style.md
+++ b/src/concepts/sql_style.md
@@ -260,7 +260,7 @@ LIMIT 10
 
 ## Join Conditions
 
-The `ON` or `USING` keyword should start on a new line indented one level more than the join keyword
+The `ON` and `USING` keywords should start on a new line indented one level more than the join keyword
 and be followed by the join conditions starting on the same line. For example:
 
 **Good**:
--- a/src/concepts/sql_style.md
+++ b/src/concepts/sql_style.md
@@ -258,6 +258,44 @@ WHERE submission_date > '20180101'
 LIMIT 10
 ```
 
+## Join Conditions
+
+The `ON` or `USING` keyword should start on a new line indented one level more than the join keyword
+and be followed by the join conditions starting on the same line. For example:
+
+**Good**:
+
+```sql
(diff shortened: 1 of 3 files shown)


relud (approved):

src/concepts/sql_style.md
```diff
@@ -258,6 +258,44 @@ WHERE submission_date > '20180101'
 LIMIT 10
 ```
 
+## Join Conditions
+
+The `ON` or `USING` keyword should start on a new line indented one level more than the join keyword
```
lucia-vargas-a, line 263:
```suggestion
The `ON` and `USING` keywords should start on a new line indented one level more than the join keyword
```

src/concepts/sql_style.md
```diff
@@ -258,6 +258,44 @@ WHERE submission_date > '20180101'
...
+and be followed by the join conditions starting on the same line. For example:
```
lucia-vargas-a, line 264: Is this how it'd be expected?
```suggestion
and be followed by the join conditions starting on the same line. The expectation is one condition per line.
For example:
```

lucia-vargas-a (approved):

src/concepts/sql_style.md
sean-rose: It depends on the nature of the join conditions, particularly whether they're wrapped in parentheses because `bqetl format` will try to fit parenthetical expressions on one line if it can.  `USING` join conditions always have parentheses and will usually fit on one line (but not always).  And in mozilla/bigquery-etl#4223 I count 37 `ON` conditions which have parentheses.

I think the "Boolean at the Beginning of Line" section already implies that compound boolean expressions should be split across multiple lines, and I was hoping to keep this section relatively simple, focusing on the positioning of the `ON` and `USING` keywords and not trying to show examples of all the possible permutations of join condition indentation that could result (which `bqetl format` will generally handle for us), so I'm personally inclined to leave this as is.

PR #800, opened 2023-11-21
GLAM artifact docs

scholtzan:
Some docs I found useful when familiarizing myself with the GLAM ETL


edugfilho (approved):
Thanks very much!

PR #801, opened 2023-11-21
Add column descriptions for urlbar events table

dzeber:
Adding column descriptions to DTMO since using `schema.yaml` in `bigquery-etl` is not currently well-supported for generated SQL (`sql_generators`).


rebecca-burwei (approved):
looks great!

PR #803, opened 2023-12-06
Add doc page for serp events

dzeber:
Adding a page describing the `serp` events themselves and the `serp_events` table.

I wrote 1 table in raw HTML so that I could get row headers spanning multiple rows.


pissac17 (approved):
LGTM

badboy:
rebased to avoid merge conflict.

PR #814, opened 2024-01-30
Update DAU definition to point to Confluence

bochocki:
This saves us from having to update the definition in two places.


jmsilverman (approved):
LGTM!!

PR #823, opened 2024-04-02
Fix typo in metric_hub.md

jmsilverman:
I *think* there's a typo in this documentation file and I *think* my new version is correct...


scholtzan (approved):
yup you are right

PR #824, opened 2024-05-01
Update docs for v2 serp_events table

dzeber:


src/datasets/other/serp_events/reference.md
```diff
@@ -430,73 +432,108 @@ Number of SERP impressions with ads loaded and an ad blocker in use:
...
+| `is_shopping_page`                           | Is the SERP a shopping page, resulting from clicking on the "Shopping" tab?                                                                                                                                                                                  |
+| `is_private`                                 | Was the SERP loaded while in Private Browsing Mode?                                                                                                                                                                                                          |
+| `is_signed_in`                               | Was the SERP loaded while signed into a search provider account?                                                                                                                                                                                             |
```
jaredsnyder, line 505: I think it'd be clearer if these were statements rather than questions, IE for `is_shopping_page` something like "True when the SERP is a shopping page, resulting from clicking the "Shopping" tab".

src/datasets/other/serp_events/reference.md
```diff
@@ -430,73 +432,108 @@ Number of SERP impressions with ads loaded and an ad blocker in use:
...
+| `ad_components`                              | Array with 1 entry for each ad component which had either an impression or engagement, `[]` if none.                                                                                                                                                         |
+| `ad_components.component`                    | SERP display component containing ad links [e.g. `ad_link`, `ad_carousel`].                                                                                                                                                                                  |
+| `ad_components.num_loaded`                   | Number of ads loaded in the component. They may or may not be visible on the page.                                                                                                                                                                           |
```
jaredsnyder, line 512: Maybe clarify that they may or may not be visible due to ad blocker or the behavior of the component

src/datasets/other/serp_events/reference.md
```diff
@@ -430,73 +432,108 @@ Number of SERP impressions with ads loaded and an ad blocker in use:
...
+| `ad_components.num_clicks`                   | Number of clicks on ad links in the component.                                                                                                                                                                                                               |
+| `ad_components.num_other_engagements`        | Number of engagements in the component which are not ad clicks. E.g. clicking "Expand" for the carousel.                                                                                                                                                     |
+| `ad_components.blocker_inferred`             | Can we infer an ad blocker is in use in the component? `true` if all loaded ads are blocked, `false` otherwise. Note that the same SERP impression can have `blocker_inferred = true` for some ad components and `false` for others.                         |
```
jaredsnyder, line 518: Turn first sentence into a statement rather than a question, could even just remove it

src/datasets/other/serp_events/reference.md
```diff
@@ -430,73 +432,108 @@ Number of SERP impressions with ads loaded and an ad blocker in use:
...
+| `num_ads_blocked`                            | Total number of ads blocked by an ad blocker in ad components for the SERP page load.                                                                                                                                                                        |
+| `num_ads_notshowing`                         | Total number of ads which are loaded but not visible, and not blocked by an ad blocker, for the SERP page load. For example, ads in the carousel that will be shown on clicking "Expand" button. Use this to count "ads that are available but not visible". |
+| `ad_blocker_inferred`                        | Can we infer an ad blocker is in use on the SERP? `true` if all loaded ads are blocked in at least one ad component, `false` otherwise.                                                                                                                      |
```
jaredsnyder, line 536: Remove the first sentence or turn into a statement

src/datasets/other/serp_events/reference.md
```diff
@@ -273,53 +285,43 @@ At the SERP level, ad ad blocker is inferred to be in use if is it inferred on a
...
+
+Tagging components as containing ads,
+and computing related fields such as `ad_components` and `num_ad_clicks`,
+is implemented at the
+[view layer](https://github.com/mozilla/bigquery-etl/blob/main/sql_generators/serp_events_v2/templates/view.sql)
+using the
+[`is_ad_component` UDF](https://github.com/mozilla/bigquery-etl/blob/main/sql/mozfun/serp_events/is_ad_component/udf.sql).
+
+The underlying
+[derived table](https://github.com/mozilla/bigquery-etl/blob/main/sql_generators/serp_events_v2/templates/desktop_query.sql)
```
jaredsnyder, line 320: These links don't work but it might just be because something isn't merged into main

src/datasets/other/serp_events/reference.md
```diff
@@ -273,53 +285,43 @@ At the SERP level, ad ad blocker is inferred to be in use if is it inferred on a
...
+instead has 2 array-valued fields:
+
+- `component_impressions`, with 1 entry per component reported in a `serp.ad_impression` event
+- `engagements`, with 1 entry per component and engagement action reported in a `serp.engagement` event.
```
jaredsnyder, line 324: This might be obvious if the links were working but I don't really understand what this means

jaredsnyder (changes requested):
Looks great!  I ran all the example queries on `mozdata.analysis.dzeber_serp_v2` and they worked great.  A few issues but they're all style things, and a couple clarifying questions

src/datasets/other/serp_events/reference.md
dzeber: For each impression we get 0 or more of each of ad_impression and engagement events. For each impression_id, the contents of these multiple events are aggregated into these arrays. The distinction is that the derived table doesn't know about ads vs. non-ads, so it's just impressions and engagements, whereas the view has ads broken out separately.

I can add more about this in the doc if you think it's helpful.

jaredsnyder: Okay that makes more sense.  It couldn't hurt to add a little more detail

jaredsnyder (approved):
Looks great!

PR #826, opened 2024-05-14
Update metric hub docs with joins and wildcard information

scholtzan:
Depends on https://github.com/mozilla/metric-config-parser/pull/366 and https://github.com/mozilla/metric-config-parser/pull/367

### DIFF
--- a/src/concepts/metric_hub.md
+++ b/src/concepts/metric_hub.md
@@ -82,6 +82,27 @@ from_expression = """
 submission_date_column = "submission_date"
 ```
 
+Data sources can be joined with other data sources:
+
+```toml
+# Join the `baseline` data source with the `metrics` data source.
+# Definitions for both data sources must exist.
+[data_sources.baseline.joins.metrics]
+relationship = "many_to_many"  # this determines the type of JOIN used; options: many_to_many, one_to_one, one_to_many, many_to_one
+on_expression = """  # SQL expression specifying the JOIN condition
+  baseline.client_id = metrics.client_id AND
+  baseline.submission_date = metrics.submission_date
+"""
+```
+
+Wildcard character can be used to apply joins to multiple data sources:
+
+```toml
+# Apply join to all data sources prefixed with user_
+[data_sources.user_'*'.joins.metrics]
+# defaults are many_to_many relationship and joins on the client_id_column and submission_date_column
+```
+
 ### `[metrics]` Section
 
 The metrics sections allows to specify metrics. A metric aggregates data and is associated with some data source.
@@ -143,6 +164,16 @@ Different statistics are available for different tools. To specify which statist
 [metrics.memory_pressure_count.statistics]
 client_count = {}
 mean = {}
+```
+
+Wildcard expressions can be used to express that a specific statistic should be available for multiple metrics:
+
+```toml
+# All metrics with the bookmark_ prefix should have the mean computed
+[metrics.bookmark_'*'.statistics.mean]
+
+# All metrics should have client counts computed (not recommended to apply statistic to every metric)
+[metrics.'*'.statistics.client_count]
 ```
 
 New statistics need to be implemented inside the tooling that uses metric definitions.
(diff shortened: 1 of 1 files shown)


src/concepts/metric_hub.md
```diff
@@ -82,6 +82,27 @@ from_expression = """
...
+Data sources can be joined with other data sources:
+
+```toml
+# Join the `baseline` data source with the `metrics` data source.
+# Definitions for both data sources must exist.
+[data_sources.baseline.joins.metrics]
+relationship = "many_to_many"  # this determines the type of JOIN used; options: many_to_many, one_to_one, one_to_many, many_to_one
+on_expression = """  # SQL expression specifying the JOIN condition
+  baseline.client_id = metrics.client_id AND
+  baseline.submission_date = metrics.submission_date
```
mikewilli, line 94: We should make sure that the default behavior is explained somewhere for each of these fields (i.e., how does the join work if they are not defined).

scholtzan: I added some more context and a clearer example of the default values to the `join` expression below this one

src/concepts/metric_hub.md
```diff
@@ -316,6 +347,32 @@ Statistics on top of these metrics need to be defined in the [`looker/` folder i
...
+    FROM
+        mozdata.telemetry.clients_daily
+"""
+columns_as_dimensions = true  # expose the selected fields as dimensions in Looker
+
+# Join `looker_base_fields` on to all the data sources in the file
+# The selected fields in `looker_base_fields` will show up as dimensions for all the metrics
+[data_sources.'*'.joins.looker_base_fields]
+
+# Overwrite the join, to allow for a different data source to be used as base field data source
```
mikewilli, line 374: It seems to me that this would not overwrite the join `data_sources.baseline.joins.looker_base_fields`, but instead add another join in addition to the one that was defined above via the wildcard join to `looker_base_fields` (`data_sources.baseline.joins.looker_base_fields`). Based on the comment, I'm guessing my understanding is off?

src/concepts/metric_hub.md
```diff
@@ -316,6 +351,32 @@ Statistics on top of these metrics need to be defined in the [`looker/` folder i
...
+        client_id,
+        os,
+        country,
+        channel
+    FROM
+        mozdata.telemetry.clients_daily
+"""
+columns_as_dimensions = true  # expose the selected fields as dimensions in Looker
+
+# Join `looker_base_fields` on to all the data sources in the file
```
mikewilli, line 370: So the wildcard scope is limited to the file, as opposed to everything that would get resolved for the current app?

scholtzan: It's limited to the current app actually. So `[data_sources.'*']` would also apply the configuration that come from other config files related to the app that are in scope.

src/concepts/metric_hub.md
scholtzan: This will actually overwrite the `joins` for the `baseline` data source to only join on `some_other_datasource`. I figured overwriting makes sense when working with wildcards, otherwise it's only possible to add more joins, but not actually remove/replace them

src/concepts/metric_hub.md
mikewilli: Ok, makes sense, that's what I expected. Thanks for clarifying! Maybe we can edit the comment to something like this?
```suggestion
# Join `looker_base_fields` on to all the data sources that are in scope for the current file (i.e., data sources for the current application)
```

src/concepts/metric_hub.md
mikewilli: Hm, yea that makes sense in terms of functionality, even if it's perhaps a bit confusing to remember that sometimes it's an addition and sometimes an overwrite, depending on whether wildcards are involved. I agree with your decision though, so we probably just want to make it extra clear in the docs.

mikewilli (approved):
This looks good to me. Do you want to run it by anyone else that might be using these capabilities? Not sure if anyone in DS or other DEs might be interested?

PR #827, opened 2024-05-24
Incident note: Spike in clients with null default search engine

pissac17:
Adding update regarding the data incident of seeing null values for default search engines for Firefox iOS


badboy:
I merged #829, which leads to a merge conflict here. Let me know if I should help with fixing that.

PR #847, opened 2024-09-24
DENG 4918 Docs update for autogenerated data checks

lucia-vargas-a:
Updated docs after implementing [DENG-4918](https://mozilla-hub.atlassian.net/browse/DENG-4918).


src/cookbooks/data_modeling/shredder_mitigation.md
```diff
@@ -165,10 +158,28 @@ We need these changes:
 
 ## Validations
 
-Recommended data validations include:
+##### Automated validations
```
BenWu, line 161: Bit of a nit and subjective but the five #'s make the text smaller than regular text so I would recommend going with the ### so it stands out.  Also applies to the section below and there are some sections above with #####
```suggestion
### Automated validations
```

BenWu (approved):

PR #859, opened 2025-02-13
Add possible values for normalized_channel and normalized_os fields

willdurand:
As discussed with @chutten, this PR adds possible values for the `normalized_channel` and `normalized_os` fields.


badboy (approved):

badboy:
Link-check failure is not your fault. I can fix that separately.

PR #903, opened 2026-07-20
Fix outdated ingestion pipeline docs

akkomar:
This corrects outdated statements in the ingestion pipeline docs, found by cross-checking each doc against the current implementation in https://github.com/mozilla/gcp-ingestion and IaC repos.


edugfilho (approved):
Thank you!

PR #904, opened 2026-07-27
Clarify live vs stable ping table docs

akkomar:
- standardize on "stable" over "historical" terminology
- fix live-table load path description (streaming path isn't used in prod)
- add current duplicate rates; drop the obsolete 0.5% AWS backfill note


BenWu (approved):
I thought duplication rates were ~5% so I'm surprised they're actually that low

